Issue #1775, opened 2014-09-12
Media box on vis page should should not have featured image label


bdonald25:
**General description:** Media box on vis pages has label for featured image, but you can't feature an image for a vis

**live/dev/localhost:** all
**iSENSE Version:** v5.6-71-g5b637f6
**Logged in (Y or N):** n/a
**Admin (Y or N):** n/a

**OS:** Mac
**Browser/Version:** Chrome

**Steps to Reproduce:** Upload an image to a project then go to vis page.

**Associated Image(s):**

![screen shot 2014-09-12 at 1 43 16 pm](https://cloud.githubusercontent.com/assets/4121924/4253796/6e3194a2-3aa4-11e4-91a3-5b521636dbac.png)

PR #1840, opened 2014-11-07
Hiding featured media label on visualizations

kcarcia:
I hid the featured media label on visualizations since imaginemagick is not working for me, and I was unable to finish allowing visualizations to have featured images. This is a temporary fix until I get that gem working again.

Towards #1775.


jaypoulz:
@stowellm You're undoing this

Issue #1890, opened 2014-12-12
Project with no title cannot be deleted


jaypoulz:
**General description:** Projects with no title cannot be deleted.

**live/dev/localhost:** live
**iSENSE Version:** v5.12
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** X11
**Browser/Version:** Firefox 34.0

**Steps to Reproduce:** 
1. Create a project titled `<script>alert(1);</script>`
2. Be amazed as the project is created, but your title is left blank.
3. Try to delete the project.
4. Weep as it doesn't delete because it has no name.

kcarcia:
TESTERS: Follow those exact steps listed above to test.

asalani93:
PASSED testing on Chrome 39.0.2171.71 (Ubuntu 14.04.1).

asalani93:
projects still aren't deleted

kcarcia:
PASSED testing on Chrome 29.0.1547.57 (mac).

kcarcia:
@asalani93 what do you mean by "projects still aren't deleted"? Did you mean to check FAILED on this one?

asalani93:
Sorry, didn't word that right.  Projects with no title are deleted.

You can still access a project by its ID - none of the projects seem to
actually be deleted, just hidden.  I'm not sure if that's in the scope of
Mike's PR or not.

kcarcia:
@asalani93 Will look into this. Nice catch!

stowellm:
What should be patched is that deleting a project from the project edit or user page should actually remove the project.  What is *not* addressed by the patch is deleting a user also deleting his or her respective projects.  That latter should be addressed in the much-talked-about-but-not-happening-yet patch where we actually delete everything on iSENSE when users choose to delete stuff (i.e. projects, data sets, media, the users themselves, etc.).

@asalani93 - when I delete a project as I'm mentioned in the former piece of the above comment, I cannot reference them again by ID (I just get an exploding beaker guy - which is good, because nothing should live at that URL anymore).  Even when I try accessing the just-deleted project via a `Project.find(#ID_OF_DELETED_PROJ#)` in a `rails c` REPL, I get an `ActiveRecord::RecordNotFound` (which is again, appropriate).  If you deleted a project through the edit or user's project page, how were you able to reference it again?

stowellm:
PASSED testing on Chrome 40.0.2214.93 (OSX Yosemite 10.10.1).

[...]

Issue #1773, opened 2014-09-12
Deleting users doesn't


asalani93:
**General description:**
Deleting a user doesn't really do as advertised.  It removes projects and saved vises as it should, but:
* you can still sign in using the garbage email address it changes your user to
* you can change your email to something else to effectively become undeleted
* you don't even need to know your deleted email to do the above if you're already signed in
* you can make projects and saved vises after you've been deleted
* you can make a saved vis for a project that has already been deleted

**live/dev/localhost:** live/dev
**iSENSE Version:** 5.6
**Logged in (Y or N):** Y and N
**Admin (Y or N):** Kind of?

**OS:** all
**Browser/Version:** all

**Steps to Reproduce:**
1. log in with a user
2. have someone delete said user while still logged in
3. go into your account settings and change your email to what it was
4. create some projects or whatever
5. things still work

-alternatively-
1. delete a user you've created
2. log out and log in to the deleted user, using the weird garbled stuff you get when you delete a user as the email
3. do things as stated above

**Associated Image(s):** N/A

kcarcia:
TESTERS: Verify you can no longer do the following:

- you can still sign in using the garbage email address it changes your user to
- you can change your email to something else to effectively become undeleted
- you don't even need to know your deleted email to do the above if you're already signed in
- you can make projects and saved vises after you've been deleted
- you can make a saved vis for a project that has already been deleted
(taken from above, written by anthony)

As a reminder, here's how this works:
"You still can't delete a project with data sets by default. If you delete a user, all of his/her data sets are deleted first. If it is found that now his/her project can be deleted (i.e. only that user contributed to it), it gets deleted. Otherwise, since that project has contributions from other users, ownership switches to admin. Then the user gets deleted. We should note only admins can delete users though."

stowellm:
PASSED testing on Chrome 42.0.2311.135 (linux).
DISREGARD THIS IS JUST A TEST

JKinsman:
PASSED testing on Chrome 42.0.2311.135 (Mac OSX 10.10.2).

jdaigneau:
PASSED testing on Chrome 42.0.2311.90 (mac).

PR #1911, opened 2015-01-09
Deleting projects actually deletes projects

stowellm:
And you can delete a project with no title :cake: 
Addresses #1890


asalani93:
Test

stowellm:
DO NOT PULL YET:
To fix:
- [ ] For saved vises: in addition to checking if project is nil, check if project just doesn't exist, then set the label to "Project: deleted" (as does currently)
- [ ] Find out why on dev the "Delete X" button won't appear for a project with data sets but on my localhost it does

stowellm:
I broke something again :( hold up.

stowellm:
@jaypoulz We've been pulling things despite not currently having an HTML validator.  Is this good to pull as well?

jaypoulz:
Not yet, I still have to test it manually

stowellm:
How's that testing going :smile:

jaypoulz:
Nice patch!

kcarcia:
@stowellm does this also address #1773?

stowellm:
@kcarcia No, this patch is strictly for deleting projects.  I've left the "deletion" (AKA hiding) of users/vis/etc alone.  I think that if we want to delete everything instead of hide everything, we should create a separate issue and patch for that purpose, especially since deleting things is bound to break a lot of tests.

kcarcia:
@stowellm derp derp derp. I missed that one major detail. Sorry about that!

Issue #1915, opened 2015-01-12
Add a key to a project via api


bdonald25:
General description: An api function to add a key to a project.

live/dev/localhost: NA
iSENSE Version: v5.8-238-g084b967
Logged in (Y or N): NA
Admin (Y or N): NA

OS: NA
Browser/Version: NA

Steps to Reproduce: NA

kcarcia:
TESTERS: Don't worry about this.

PR #1916, opened 2015-01-12
API function for adding a key to a project

bdonald25:
@jaypoulz 
#1915
I still need to update the api docs. Could you point me in the right direction?


jaypoulz:
Alright, so the file you are looking for is located at: `app/views/home/api_v1.html.erb`
It's really just a giant html doc, so nothing complicated there.

bdonald25:
Any idea why this failed? It says upload media failed but I didn't touch anything relating to media.

JKinsman:
Sorry @bdonald25 !  I "fixed" that test with my most recent patch.  It wasn't failing when I ran it locally or on Travis, but apparently the test still fails sporadically.  I'll take a look at it tomorrow morning.  In the mean time, try rerunning the test on Travis again.

bdonald25:
@JKinsman Ok, it passed when I reran it. Thanks!

app/controllers/api/v1/projects_controller.rb
```diff
@@ -36,7 +36,35 @@ def create
           end
         end
       end
+
+      def key
```
jaypoulz: would you mind renaming this to add_key or create_key. I feel like it would me slightly more clear

app/controllers/api/v1/projects_controller.rb
```diff
@@ -36,7 +36,35 @@ def create
...
+        @project = Project.find(params[:contrib_key][:project_id])
+
+        if @cur_user.id == @project.user_id
+          session[:name] = params[:key_name]
+          session[:key] = params[:key]
+          session[:project_id] = @project.id
+          key = ContribKey.new(contrib_key_params)
+          key.save
+        else
+          errors = 'User does not own the project.'
```
jaypoulz: Can this error block go in the do | format | below so that we have one less else block? It's checking the same condition below after all.

test/integration/api_v1_test.rb
```diff
@@ -586,6 +586,27 @@ class ApiV1Test < ActionDispatch::IntegrationTest
     assert_response :unauthorized
   end
 
+  test 'add a key to a project' do
```
jaypoulz: Would you mind adding tests for (at least) the forbidden case as well? It would also be nice to have a test for the unprocessable variety - maybe if you name the params wrong that will trigger?

bdonald25: @jaypoulz Is there a list somewhere of other users I can use for tests? The only user I could find in this test file is kcarcia@cs.uml.edu and I am not sure where that user lives.

jaypoulz: Yes haha. So the other users are in test/fixtures/users.yml :rabbit:

bdonald25:
@jaypoulz

test/integration/api_v1_test.rb
bdonald25: I tried to write a test for unprocessable using the wrong params but the test gave an error instead.

jaypoulz: Actually, the error is a good thing!!! Whenever you have a test that fails, you should modify your code so that it passes your test. This is called test-driven development. It's super popular these days (at startups especially), so this is a good learning experience for the both of us. So let's have a look:
```
@project = Project.find(params[:contrib_key][:project_id])

if @cur_user.id == @project.user_id
  session[:name] = params[:key_name]
  session[:key] = params[:key]
  session[:project_id] = @project.id
  key = ContribKey.new(contrib_key_params)
  key.save
  end
```
First things first: you must check if the things in params are nil!!!
This means that params[:contrib_key], params[:contrib_key][:project_id], params[:key_name], and params[:key] must all be checked non-nil.

Since we are special casing success, I would skip the if block if any of these values are nil. This way the respond to |format| block can remain the way it is.

Otherwise this patch is quite elegant.

bdonald25:
@jaypoulz I am getting "test/integration/api_v1_test.rb:3:1: C: Class definition is too long. [506/500]" from travis. Does this mean I have to put the test in another file?

jaypoulz:
Nah this is a rubocop thing.

All you need to do is edit rubocop.yml (it's in the base directory) and up the class-length cap from 500 to, maybe 600? Up to you on the new cap. I feel like the API tests belong in the same file. The only issue will be if that file gets really long. For now lets just raise the cap.

bdonald25:
@jaypoulz

jaypoulz:
Checking each individual nil-ity and reporting a specific error is a little overkill, and your implementation is not the most minimal, but I'll let it slide because I dig user feedback. Nice patch!

Issue #1659, opened 2014-07-25
Update contact info for new employees


AlanRosenthal:
**General description:**
http://isenseproject.org/contact

**live/dev/localhost:** N/a
**iSENSE Version:** N/a
**Logged in (Y or N):** N/a
**Admin (Y or N):** N/a

**OS:** All
**Browser/Version:** All

**Steps to Reproduce:** N/a

**Associated Image(s):** None

kcarcia:
TESTERS: Tell us what you like and don't like. Are there typos? How does the page work responsively?

asalani93:
PASSED testing on Chrome 39.0.2171.71 (Ubuntu 14.04.1).
My linked in is wrong:
https://www.linkedin.com/in/asalani

kcarcia:
FAILED testing on Chrome 29.0.1547.57 (mac).
General:
There are inconsistencies:
- Either all degrees should have a comma before the year of not. Mine is like "B.S. Computer Science 2015" while Jeremy's is like "M.S. Computer Science, 2016" Seems like most have the comma, so I'd stick with that.
- With the exception of Nat, I'd put all emails in the blahblahblah AT blahblahblah format.
- It might not be necessary to say student after undergraduate or graduate. Really this is up to you. If you keep it as is, I would add Student to Hassan.

Fred:
Director Student Success -- should this say Director of Student Success?

Jeremy:
Should we give him a more specific manager title? Lab manager? Project manger?

Anthony:
Anthony's linked in link doesn't work.

stowellm:
Thanks for all the feedback, @kcarcia!  Definitely caught some things I missed.  All of the above should now be addressed and is in the pending PR, #1957.  I also got a chance to add Skittles to the contact page as well.

kcarcia:
@stowellm Awesome! Will review first thing when I get in tomorrow!

asalani93:
PASSED testing on Chrome 39.0.2171.71 (UUBBUUNNTTUU 1144..0044).

bdonald25:
PASSED testing on Chrome 41.0.2272.35 (mac).

stowellm:
PASSED testing on Chrome 40.0.2214.93 (OSX).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).
Looks good!  @stowellm if you could add 2015 as my year of graduation, I'd greatly appreciate it! :)

PR #1926, opened 2015-01-16
New Contact Page

stowellm:
Addresses #1659 

Go ahead and read the HTML if you really want to.  But here's some screenshots to check out.

---

1/3

contact1

---

2/3

contact2

---

3/3

contact3


JKinsman:
@stowellm this passes my review.  Everything looks good! :smile:

jaypoulz:
Wait for Skittles

stowellm:
@jaypoulz  - should we pull this now rather than wait?  If the people we contacted truly wanted to be on the page, they would have responded. At what point are we more or less forcing them onto the page because we want them there as opposed to them willingly being there?

kcarcia:
@stowellm do we assert the HTML is valid on this page? If so, please mark as reviewed. Otherwise, are you able to add a test that does so please?

stowellm:
@kcarcia Yes, the contact page HTML is checked.

`home_controller_test.rb` lines 16-20:
```
  test 'should get contact' do
    get :contact
    assert_response :success
    assert_valid_html response.body
  end
```

Going to merge with upstream and mark as reviewed as requested.

Issue #1829, opened 2014-10-23
Media block shows up on project vis page


fgmart:
**General description:**
Why does the 'Media' block show up on project vis page? I don't understand why that's there.

**live/dev/localhost:** Live
**iSENSE Version:** N/a
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** N/a
**Browser/Version:** N/a

**Steps to Reproduce:**
Got a project's vis page.

**Associated Image(s):** N/a

kcarcia:
@fgmart The current way to add photos to the photo viz is through this media box. If we every re-work photo vis, we can determine if there's a better to prompt users for photos.

jaypoulz:
I love the idea of uploading photos via the photos vis. @stowellm Feel up to it?

PR #1929, opened 2015-01-21
Removed media block from project's vis pages

kcarcia:
When you click an individual data set in a project, the media block appears on the vis's page. it doesn't make sense to have that there since it does not appear when you are visualizing all data sets.

The media block still appears for saved visualizations, though.

Addresses #1829.

### DIFF
--- a/app/views/visualizations/displayVis.html.erb
+++ b/app/views/visualizations/displayVis.html.erb
@@ -72,13 +72,3 @@
     </div>
   </div>
 </div>
-<div class="row">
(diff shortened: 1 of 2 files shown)


stowellm:
Closing this pull request - currently, the only way to add a picture to a single data set is by using this media box.  Without the media box, we longer can add photos to data sets, thus rendering the photo vis entirely useless.  We will have to come up with an alternative to adding photos to a data set if we are to remove this media box.  @kcarcia will be handling this either in a new issue or via discussion with others

PR #1933, opened 2015-01-22
HTML validation tests should regardless of 3rd party status

jaypoulz:
I also refactored the pages that were failing html validation, especially those whose tests had be commented out.


asalani93:
My only comment is that `<br/>` is only for adding line breaks to text, and not for formatting purposes.  However, that's a site-wide problem and fixing it on just the files you touched seems a bit silly and tangential to the original purpose of the PR, so IDGAF.

That should be fixed, but not here I guess.

Issue #1867, opened 2014-12-08
Bug Reporting Bug Fixes for v2


kcarcia:
**General description:** The following issues should be addressed as part of v2 of Bug Reporting:
- [x] <strike>@cur_user should only be set once. (Said by Jim)</strike> Update: this is where it's set.
- [x]  Most of the Authenticator stuff should be in lib/github_authenticator.rb or similar, not directly in the controller. (Said by Jim)
- [x] Increase test coverage back to 90.
- [x] It should not say anonymous user.

**live/dev/localhost:** live
**iSENSE Version:** v5.8-220-g875ca7e
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 39.0.2171.71

**Steps to Reproduce:** You can access bug reporting from the footer.

kcarcia:
TESTERS: Bug reporting on dev should work with and without a GitHub account.

asalani93:
PASSED testing on Chrome 39.0.2171.71 (Ubuntu 14.04.1).

kcarcia:
PASSED testing on Chrome 29.0.1547.57 (mac).

stowellm:
PASSED testing on Chrome 40.0.2214.93 (OSX Yosemite 10.10.1).

asalani93:
PASSED testing on Chrome 39.0.2171.71 (UUBBUUNNTTUU 1144..0044).

bdonald25:
PASSED testing on Chrome 41.0.2272.35 (mac).

stowellm:
PASSED testing on Chrome 40.0.2214.93 (OSX).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

PR #1936, opened 2015-01-24
Bug reporting fixes for v2

kcarcia:
Ok, so I grabbed Jim's pulled request (which added some tests and moved authenticator stuff to the lib) to fix the failing aspects. Then, I also

1) Upped test coverage to 90%
2) Removed any external notation of users being "anonymous" when reporting issues.

Addresses #1867


jaypoulz:
My only comment on this patch - which is extremely elegant, by the way, is that if it doesn't already do so, the report a bug pages (on our side) should undergo html validation. I'm pretty sure this is occuring for the actual report a bug page (with the links to github or anonymous) but I don't know if it occurs for the bug description submission. The only problem with this is that I don't think you can reach that page without going through the github authentication first. Would it be possible to run html validation explicitly on this file?

kcarcia:
@jaypoulz it definitely validates the report a bug page. I actually had to fix some of the HTML on that page because it was invalid. But you're right. You can't get to the other pages without having those env variables. I will merge this for now but just made note to talk to you on Monday about getting those other pages to validate.

Issue #1904, opened 2015-01-05
API Posting Dataset 


tylerpuleo:
**General description:** When posting datasets with the API the first field sets the size of the dataset so if the second field is longer it gets truncated.

**live/dev/localhost:** dev
**iSENSE Version:** v5.8-238-g084b967
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu 14.04
**Browser/Version:** Chrome 39.0.2171.95

**Steps to Reproduce:** Make a post request to a project where the second field is longer.

tylerpuleo:
@bdonald25

kcarcia:
TESTERS: Don't worry about this.

PR #1943, opened 2015-01-26
Data is no longer lost when uploading through the api

bdonald25:
#1904
@jaypoulz


jaypoulz:
Couple things left to do here:

1. Add a test
2. Remove the puts lines

bdonald25:
@jaypoulz

jaypoulz:
@bdonald25 for future reference, you went overboard on the test again. Just one of the asserts at the end would have been sufficient. (Maybe just asserting if the last number was there). That being said, this patch is clean. Nice work! :smile_cat:

Issue #1909, opened 2015-01-08
Exponential regressions are broken.


JKinsman:
**General description:** Trying to create an exponential regression results in an alert box displaying an error message that the model was unable to create an exponential model for the data, even when the data is inherently exponential in nature.

**live/dev/localhost:** live
**iSENSE Version:** v5.12
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** X11
**Browser/Version:** Chrome 38.0.2125.101

**Steps to Reproduce:** 

315

kcarcia:
TESTERS: @JKinsman worked his math magic to fix linear, quadratic, cubic, exponential, and logarithmic regressions on both Timeline and Scatter. To test this and #1299, please test the various types of regressions.

stowellm:
PASSED testing on Internet Explorer 11.0 (windows 7).
Mostly works - I ran into one instance I couldn't draw an exponential but can't seem to reproduce it now.

asalani93:
PASSED testing on Chrome 39.0.2171.71 (Ubuntu 14.04.1 x64).

tylerpuleo:
PASSED testing on Chrome 38.0.2125.111 (linux).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

Issue #1299, opened 2014-01-24
Cubic regression best fit on timeline not working


kcarcia:
Specifically occurred on this vis last night at the workshop: http://isenseproject.org/visualizations/208

@jaypoulz mentioned something sneaky may be going on with the math, @mmcguinn.

kcarcia:
TESTERS: See #1909 for testing this. You can steal the data from the project in the description to test the cubic regression.

stowellm:
PASSED testing on Internet Explorer 11.0 (windows 7).
Cubic works, but log isn't.  See http://rsense-dev.cs.uml.edu/visualizations/449

asalani93:
PASSED testing on Chrome 39.0.2171.71 (Ubuntu 14.04.1 x64).

tylerpuleo:
PASSED testing on Chrome 38.0.2125.111 (linux).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

PR #1955, opened 2015-01-27
Full set of regressions for Timeline and Scatter 

JKinsman:
Addresses issues #1909 , and #1299 .

Linear, Quadratic, Cubic, Exponential, and Logarithmic regressions now work properly on both Timeline and Scatter.

Change summary:

1.  New normalization strategy
2.  New form of exponential and logarithmic hypotheses
       -The old logarithmic function was nonlinear in the parameters, which broke our regression model
       -Changed exponential to match, as the basis (either the log function, or the exponential function), must be    
        applied to the entire equation, not just parts of it
3.  Fixed a few partial derivatives
     -Some were just plain wrong, and some had to be adjusted for the new hypotheses
4.  Recovering the original parameters of the regression
     -When data normalization was added, the regression model trained on the normalized data.  The original           
      parameters (displayed in the tooltip), were being computed incorrectly.  The new normalization strategy 
      computes the "real" parameters of the regression by solving a system of linear equations, without the need 
     for denormalized functions, standard deviation, etc etc.

@jaypoulz


app/assets/javascripts/visualizations/highvis/regression.coffee
```diff
@@ -364,12 +319,78 @@ $ ->
       mean
 
     # Normalize
-    normalizeData = (points, mean, sigma) ->
-      (point - mean) / sigma for point in points
+    normalizeData = (points, type) ->
+      max = Math.max.apply(null, points)
+      min = Math.min.apply(null, points)
+      ret = if type == globals.REGRESSION.LOGARITHMIC
```
jaypoulz: would you mind bringing the if down the next line. it will end up looking like this
```
ret =
  if type == globals.REGRESSION.LOGARITHMIC then points
  else points.map((y) -> ((y - min) / (max - min)) + 1)
```

jaypoulz:
Passed code review with one minor style change. I'm going to pull this down and test it manually.

JKinsman:
I made the styling change requested :)

stowellm:
That is the longest branch name I think I've ever seen :fearful:

jaypoulz:
LOL @stowellm :smile_cat:

Issue #1952, opened 2015-01-27
Google Doc Contribution Dialogue Box Needs Some UI Love


kcarcia:
**General description:** The dialogue box used to contribute data via a Google Doc needs some UI work. See the image below.

**live/dev/localhost:** live
**iSENSE Version:** v5.12
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 40.0.2214.91

**Steps to Reproduce:**
1. Go to a project.
2. Click Google Doc to contribute data.

**Associated Images:**
![screen shot 2015-01-26 at 10 35 56 pm](https://cloud.githubusercontent.com/assets/1798956/5913100/1e36cfa8-a5ac-11e4-8772-20794a7f7997.png)

stowellm:
I was just about to make an issue about this until I saw you already made one.  What I wanted to mention is that the `Save` button should be on the right.  Convention appears to be to always put positive action buttons on the right, negative on the left.

kcarcia:
@stowellm +1 !

kcarcia:
TESTERS: Alignment on the Google Doc box should be fixed. Also, the Save should be on the right. Cancel on the left.

stowellm:
PASSED testing on Internet Explorer 11.0 (windows 7).

asalani93:
PASSED testing on Chrome 39.0.2171.71 (Ubuntu 14.04.1 x64).

tylerpuleo:
PASSED testing on Chrome 38.0.2125.111 (linux).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

jdaigneau:
PASSED testing on Chrome 40.0.2214.94 (Max OSX 10.9.4).

PR #1977, opened 2015-02-04
Aligned msg on Google Doc modal/switch order of save and cancel buttons

kcarcia:
Addresses #1952.


app/views/projects/show.html.erb
```diff
@@ -320,10 +320,10 @@
           <input name="pid" value="<%=@project.id%>" type="hidden">
           <input name="gdoc" id="doc_url" class="form-control" type="text" style="width:95%">
         </div>
-        <i style="padding-left:15px;">*This can be generated by clicking 'File->Publish to the web' or 'File->Share' on your spreadsheet.</i>
+        <i>This can be generated by clicking 'File->Publish to the web' or 'File->Share' on your&nbsp;spreadsheet.</i>
```
asalani93: why is this an `&nbsp;`?  it seems like it could just be a space

asalani93:
Aside from the `&nbsp;` thing, it works and stuff and looks good.  I don't really get why it's there.  I tried removing it and the only thing I noticed is that "your" and "spreadsheet" would sometimes be on different lines w/o it.  Is this why it was added?

kcarcia:
@asalani93 Exactly. Widows are bad!

PR #1984, opened 2015-02-06
allow data_set visualizations to be embedded too

scytacki:
this makes it possible to embed URLs like: 
/projects/[projectNum]/data_sets/[dataSetIdList]?embed=true

Being able to embed those URLs is useful so I can make a tool which filters lists of data_sets for the user and shows the default visualization of this filtered set of data_sets.

### DIFF
--- a/app/controllers/visualizations_controller.rb
+++ b/app/controllers/visualizations_controller.rb
@@ -4,7 +4,7 @@ class VisualizationsController < ApplicationController
 
   skip_before_filter :authorize, only: [:show, :displayVis, :index, :embedVis]
 
-  after_action :allow_iframe, only: [:show]
+  after_action :allow_iframe, only: [:show, :displayVis]
 
   # GET /visualizations
   # GET /visualizations.json


kcarcia:
@jaypoulz @jdalphon @bdonald25 I don't know too much about the API. This pull request really deserves your attention.

jaypoulz:
The clears my review, but Jim might have more to say than Bobby or myself on this particular issue. It's not really the API that is being changed.

The change here is making visualizations directly embeddable without saving them first. When we designed iSENSE, our end goal was to provide a tool so that users could produce visualizations and share them. This is why only 'saved vises' are embeddable at the moment - these are visualizations ready to be shared.

However, given that we've since added 'default vises' (and heavily extended the feature set for these), it may be in the spirit of the website to allow for embedded vises directly from projects. Since the scope of this change is really at the project intent/design level, I think @jdalphon  or @fgmart should have the final say.

fgmart:
@jaypoulz thank you for this explanation.

I agree: it's now reasonable for project viz's themselves to be embeddable. I think this adds a lot of flexibility to how iSENSE can be used -- specifically, you'd no longer have to publish a viz to use it in another context.

@jdalphon OK to pull?

jdalphon:
Didn't this already do what you are saying. For instance 
http://isenseproject.org/projects/106/data_sets/2801?embed=true has always worked. 

Am I missing something?

jdalphon:
@scytacki @fgmart @jaypoulz 

Oh, Sorry. I was mistaken. This would allow for external usage through an iframe. Yes this is fine, it can be pulled.

scytacki:
That link will not work if you put it in an iframe that is on another
domain. This change makes it work in cross domain iframes.

scytacki:
Thanks for merging this. @jdalphon sorry about my comment above, I saw you first message but missed your second one.

PR #1986, opened 2015-02-09
API call to check if a key exists

bdonald25:
@jaypoulz 
I know you are going to say I went overkill with the errors again, but this time I think it needed to explicitly test if it is the project or the key that doesn't exist.


app/controllers/api/v1/projects_controller.rb
```diff
@@ -37,6 +37,31 @@ def create
         end
       end
 
+      def key
+        if params[:id].nil? || params[:contribution_key].nil?
+          respond_to do |format|
```
jaypoulz: This line:
```
respond_to do |format|
```
is in every terminating condition of your code.

It would be more efficient to wrap ALL of you if statements (and else statements) inside a single one of these. That way, you only have that line once.

app/controllers/api/v1/projects_controller.rb
```diff
@@ -37,6 +37,31 @@ def create
...
+            format.json { render json: { error: 'Neither Project ID nor Contribution Key can be empty.' }, status: :unprocessable_entity }
+          end
+        else
```
jaypoulz: Instead of an 'else' here, just have a return at the end of your 'if'

This way, you don't have to indent a level inward when you go through the other conditions.

app/views/home/api_v1.html.erb
```diff
@@ -196,6 +196,48 @@
   <div class="col-md-12">
     <table class="table table-bordered api">
       <tr>
+        <td class="GET def">GET</td>
+        <td class="method">Key</td>
+        <td class="code">/api/v1/projects/XX/key/</td>
+        <td class="args">contribution_key(req)</td>
+      </tr>
+      <tr class="more" style="">
```
jaypoulz: why is style there?

app/views/home/api_v1.html.erb
```diff
@@ -196,6 +196,48 @@
...
+          <b>Response Codes</b>
+          <ul>
+            <li>Success: 302 Found</li>
+          </ul>
+          <ul>
+            <li>Fail: 404 Not Found</li>
+          </ul>
+           <b>Data</b>
+          <pre>
+  {
```
jaypoulz: do you mind adjusting the indentation here

app/views/home/api_v1.html.erb
```diff
@@ -196,6 +196,48 @@
...
+    "contribution_key" => "contribution key (string)"
+  }
+          </pre>
+          <b>Response</b>
+          <pre>
+  {
```
jaypoulz: indentation

app/views/home/api_v1.html.erb
```diff
@@ -196,6 +196,48 @@
...
+    "contribution_key": "(string)"
+  }
+          </pre>
+          <b>Notes</b>
+          <ul>
+			<li>XX should be replaced with the ID for the project.</li>
```
jaypoulz: here to - this is the curse of using tabs instead of spaces

test/integration/api_v1_test.rb
```diff
@@ -608,6 +608,20 @@ class ApiV1Test < ActionDispatch::IntegrationTest
...
   end
 
+  test 'check if key exists' do
+    pid = @dessert_project.id
+    get "/api/v1/projects/#{pid}/key",
+        contribution_key: 'apple'
+    assert_response 302
+  end
+
+  test 'check if key exists (does not exist)' do
```
jaypoulz: rename this test to be less confusing. Also, make sure you have tests for all 4 conditions defined in the function.

1. bad params
2. no project
3. no key
4. success

app/views/home/api_v1.html.erb
bdonald25: Good question. It is there for every "more" class on this page. It doesn't seem to change when I remove it. Should I remove it everywhere it exists on this page?

app/views/home/api_v1.html.erb
bdonald25: Are you referring to this?
{
    "contribution_key" => "contribution key (string)"
}

If I indent this it will be awkwardly indented on the website.

app/views/home/api_v1.html.erb
jaypoulz: Yes. An empty inline style declaration isn't contributing anything :smile:

bdonald25:
@jdalphon

Issue #1980, opened 2015-02-05
"Select All" is broken on tables


bdonald25:
**General description:** "Select all" is broken on tables.

**live/dev/localhost:** live
**iSENSE Version:** v6.0
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 41.0.2272.35

**Steps to Reproduce:** If you click a dataset to disable it and then re-enable the same dataset, the select all button will not re-enable but will act like it is enabled. (ie, when you select it, all the datasets become deselected and vice versa)

393

stowellm:
Steps to Reproduce (modified since it took me a while to understand Bobby's instructions):
1. Visualize a group of data sets with some data and multiple fields
2. Go to table
3. Make sure #Check All is selected along with the other fields
4. Deselect one field. Note #Check All is also deselected
5. Reselect that field.  Note #Check All is still deselected
6. Check off #Check All. Note the bug: all the other fields are now *deselected* and this checkbox's functionality has reversed

kcarcia:
TESTERS: Please follow the beautiful steps to reproduce @stowellm posted above.

stowellm:
PASSED testing on Internet Explorer 11.0 (windows 7).

asalani93:
PASSED testing on Chrome 39.0.2171.71 (Ubuntu 14.04.1 x64).

tylerpuleo:
PASSED testing on Chrome 38.0.2125.111 (linux).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

jdaigneau:
PASSED testing on Chrome 40.0.2214.94 (Max OSX 10.9.4).

PR #1993, opened 2015-02-10
Fixed check-all bug in table for visible fields

stowellm:
Addresses #1980 

The problems were:
- the code was checking the length of all fields versus selected fields instead of very simply checking if the "#Check All" box was selected or not, leading to the error described in the issue
- yFields was improperly being modified in the click event of the "#Check All" box - switched it to a temp variable called selFields

Additionally:
- Changed some "($ " to "$("
- Changed the name of the box to "#Check All" instead of "#Select All" to be consistent with the groups check-all box

### DIFF
--- a/app/assets/javascripts/visualizations/highvis/table.coffee
+++ b/app/assets/javascripts/visualizations/highvis/table.coffee
@@ -61,7 +61,7 @@ $ ->
 
       start: ->
         # Make table visible? (or something)
-        ($ '#' + @canvas).show()
+        $('#' + @canvas).show()
 
         # Calls update
         super()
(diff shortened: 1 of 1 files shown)


jaypoulz:
Why indeterminate? Is this style used somewhere else in the site?

stowellm:
Indeterminate because I copy/pasted this code from baseVis.  Yes, its used elsewhere.

jaypoulz:
Ahh ok. Well for grouping, it's different -> the grouping controls have a hierarchical structure to them. These do not. so we're in the clear.

Please no indeterminate here.

stowellm:
You got it, doc! :rabbit: :tomato: (<-- please settle the tomato - there are no carrots)

jaypoulz:
GG good to pull

Issue #1989, opened 2015-02-10
API Contrib Key Takes Project ID in url and as a param


tylerpuleo:
General description: Project ID listed twice in Contrib key api call "/api/v1/projects/XX/add_key"

live/dev/localhost: live
iSENSE Version: v6.0
Logged in (Y or N): Y
Admin (Y or N): N

OS: Ubuntu 14.04 (Jacascript)
Browser/Version: Chrome 41.0.2272.43

Steps to Reproduce: Look at docs. Need to pass ID twice.

jdalphon:
@bdonald25 I will consider your pull if you can answer this question. Could this, at all, cause a current working consumer of the API to now fail? Basically you have to prove that the call
`isenseproject.org/api/v1/projects/3/add_key?...  contrib_key: { project_id: something other than 3, ...} `
failed previously (error case, not just do the wrong thing)
I know this is dumb but if someone literally typed `isenseproject.org/api/v1/projects/3(or XX)/add_key...some other number` what would happen. 

Looking at the code I know the answer, but this is a good test to make sure we are not breaking things for users. Please post the output from tests. 

Extra credit (ill buy you a coffee) what happens if you make the call owning the two different projects. i.e.  `/api/v1/projects/2/add_key project_id: 3` where you own 2 and 3.

bdonald25:
@jdalphon 

Before it was using the param "project_id" not the "id" param in the url. 
/api/v1/projects/2/add_key project_id: 3  -> will currently create a key for project 3
Now it uses the project id from the url, so now it will create the key for project 2. 

The proper way I should have done it if I was worried about compatibility would be to change the url parameter to "project_id". However, I went with "id" because that is what all of the other api calls use as a parameter and this api call has only existed for 10 days so I am not worried about it breaking anything.

jaypoulz:
@bdonald25 That's fine, but let's be more mindful about this in the future. You're right that it is unlikely that anyone depends on this, but as a general rule of thumb, anything we release out live should be considered supported. Is the update for add key already live?

bdonald25:
@jaypoulz No this update is currently pending as a pull request. Keep in mind it would only have unintended consequences if someone had a different id for both project parameters. If they had previously used the same project id in the url as the one they passed as a parameter it will still work.

jaypoulz:
I left a comment in the pull request as to how to fix this without 'breaking' anything. You're are right that having both params be different is a problem, but for that we just set a default.

[...]

PR #1999, opened 2015-02-11
add key api call had two parameters for the project id

bdonald25:
#1989

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -46,7 +46,7 @@ module Api
           respond_to do |format|
             format.json { render json: { msg: 'name is nil' }, status: :unprocessable_entity }
           end
-        elsif params[:contrib_key][:project_id].nil?
+        elsif params[:id].nil?
           respond_to do |format|
             format.json { render json: { msg: 'project id is nil' }, status: :unprocessable_entity }
           end
@@ -55,7 +55,7 @@ module Api
             format.json { render json: { msg: 'key is nil' }, status: :unprocessable_entity }
           end
         else
-          @project = Project.find(params[:contrib_key][:project_id])
+          @project = Project.find(params[:id])
 
           if @cur_user.id == @project.user_id
             session[:name] = params[:key_name]
--- a/app/views/home/api_v1.html.erb
+++ b/app/views/home/api_v1.html.erb
@@ -169,7 +169,6 @@
         "email" => "email address (string)",
         "password" => "password (string)",
         "contrib_key" => {  'name' => "Key's Name(string)",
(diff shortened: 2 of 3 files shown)


jaypoulz:
I think it might be best to make the second param silently do the right think, instead of removing it altogether. That is, if [:id] OR [:project_id] exist, grab one and perform the action:

2. unless [:id].nil? id_you_should_use = [:id] else [:project_id]

bdonald25:
@jaypoulz id will always exist though as is in the url. Are you suggesting that I use :project_id as a default over :id?

Previously :id was in url and :project_id was passed as a param. :project_id was used and :id was never looked at.
Now :id is still there but is actually being used and :project_id is gone. 

Do you suggest that I use :project_id if it exists else use :id?

jaypoulz:
I am suggesting the exact opposite of what you just said.
Default to [:id]

bdonald25:
That does not fix the problem though.
before isenseproject.org/api/v1/projects/3/add_key?... contrib_key: { project_id: 2, ...} -> adds key to project 2.
now (same api call) isenseproject.org/api/v1/projects/3/add_key?... contrib_key: { project_id: 2, ...} -> adds key to project 3
If I default to :id that problem will still exist. :id still exists in both api calls.

jaypoulz:
Yes. The correct behavior for isenseproject.org/api/v1/projects/3/add_key... contrib_key: { project_id: 2, ...} is to add a key to project 3.

My confusion was that i didn't realize that the url param 'id' MUST exist because it is in the url. You don't even have to add a check for project_id, just ignore it altogether.

bdonald25:
@jaypoulz Yes I agree. That is why I asked if it should default to project_id so that it continues to post to 2.

jaypoulz:
My argument is - but that's stupid.

You could add an error case for this call that checks if they're the same, and if they aren't, spit out an error.

This way, it might break the end-user's behavior, but it would tell them why.

OR

We could default to project_id even though that's just disgusting.

jdalphon:
@bdonald25 You'll get your coffee if I am ever in the lab again. You are correct as to what it would do. @jaypoulz && @bdonald25 This is a pretty odd case. The solution to the problem depends on why add_key was added to the api in the first place. If an external user requested it then it needs a patch as Jeremy pointed out. If it was something that someone internal asked for, I would say this pull works. 

In any case, from now on I should be the only one reviewing API changes. I didnt look at the add_key (at least not well) and I will take the blame on that one. If you dont hear from me on a pull quickly just send me a hangouts message and ill get right on it.

jaypoulz:
@jdalphon Y'okay :smile:

bdonald25:
@jdalphon I was the one who wanted it. :)

Issue #1669, opened 2014-07-30
Change project title does not work in Internet Explorer


JKinsman:
**General description:**
On the project show page, Clicking the "edit title" icon next to the project's title and saving a new title will not actually change the Project Title.

**live/dev/localhost:** dev/localhost
**iSENSE Version:** 5.4-120
**Logged in (Y or N):** Y
**Admin (Y or N):** N/A

**OS:** Windows 8.1
**Browser/Version:** Internet Explorer 11

**Steps to Reproduce:**
In Internet Explorer,
1.  Create a Project
2.  Enter a title
3.  On the project Show Page, attempt to edit the title

**Associated Image(s):** None

stowellm:
@jaypoulz Can this be closed as wontfix?  No one has IE :smile:

jaypoulz:
@stowellm Nope - schools still use i.e. Don was using it at Tyngsboro high when he reproduced it just a month ago. We need to fix this if possible. Testing will require a vm.

kcarcia:
TESTERS: I will pick a designated person to test this.

stowellm:
PASSED testing on Internet Explorer 11.0 (windows 7).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

PR #2001, opened 2015-02-12
IE is bad and should feel bad.  Changing project titles works

stowellm:
Addresses #1669 
Thanks to @asalani93 for basically helping with 99% of the work here :smile_cat: 
Can't wait to reclaim *half* of my Ubuntu partition back that is currently being eaten by an IE-Win7-VM.


jaypoulz:
would you mind removing  `window.` before location.pathname? it should do the same thing

stowellm:
Por supuesto.

Issue #2009, opened 2015-02-18
Edit_fields Table v2: Should Only Submit Data on Save


kcarcia:
**General description:** Currently, the edit_table form submits every time we add a field, and ajax is used to delete fields. The form should submit only on save. Any request to the server should only be made on save. Therefore, I will be redoing the form using Coffeescript. When I am done users will be able to do the following:

- [x] Add fields.
- [x] Delete fields.
- [x] Submit the form ONLY on save.

As a result, the following UI changes will be made:

- [x] Add a cancel bottom.
- [x] New rows highlighted on save.

**live/dev/localhost:** live
**iSENSE Version:** v6.15
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** Macintosh
**Browser/Version:** Chrome 40.0.2214.111

**Steps to Reproduce:**  None.

kcarcia:
TESTERS: Go to the edit fields table, and do your best to break it! Make sure you specifically take note of timestamp and location fields—both should only be added once.

bdonald25:
PASSED testing on Chrome 42.0.2311.135 (mac).

JKinsman:
PASSED testing on Chrome 42.0.2311.135 (Mac OSX 10.10.2).

jdaigneau:
PASSED testing on Chrome 42.0.2311.90 (mac).

jdaigneau:
PASSED testing on Chrome 42.0.2311.90 (mac).

Issue #2002, opened 2015-02-13
Edit_fields UI Makeover


kcarcia:
**General description:** The Edit Fields page is getting a makeover. Changes will include the following:
- [x] Add header to page.
- [x] Buttons to add fields instead of a dropdown list.
- [x] Buttons moved to bottom of page.
- [x] Option to add more than one Number and Text field at a time.
- [x] Delete button should match delete button on Slickgrid.
- [x] Restrictions and Units columns should be hidden in mobile.
- [x] Restriction input boxes should be hidden for everything but text fields.  Unit input boxes should be hidden for everything but number fields.
- [x] Change Save button to match Save button Slickgrid.

**live/dev/localhost:** live
**iSENSE Version:** v6.0
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 40.0.2214.111

**Steps to Reproduce:** None.

kcarcia:
TESTERS: Verify everything in the list above has been accomplished.

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).
Looks great, and behaves appropriately :)

stowellm:
PASSED testing on Chrome 40.0.2214.115 (linux).
Notes:
- Using a project with some fields and data sets, I am warned I cannot delete fields.  But I can add them.  If I add a new field, I should be able to delete that field before saving it.  However, I cannot.

kcarcia:
@stowellm Nice catch, Mike. Will be addressed in my next patch. :)

jdaigneau:
FAILED testing on Chrome 40.0.2214.115 (mac).
The buttons to add fields is at the top of the page rather than the bottom, as it was intended.

kcarcia:
@jdaigneau Good catch. I should've removed that from the issue. For this particular iteration, the buttons should be at the top.

bdonald25:
FAILED testing on Chrome 40.0.2214.115 (linux).
If you minimize the width of the page too much, you lose the notes and the ability to add more than one field at a time.

kcarcia:
@bdonald25 These things are both intended. Otherwise, the page become unusable.

Issue #1966, opened 2015-01-30
After removing a location field you can't add another one.


bdonald25:
**General description:** After deleting a location field you can't add another one without saving.

**live/dev/localhost:** live
**iSENSE Version:** v5.12-208-gdd7f7af
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 41.0.2272.35

**Steps to Reproduce:** On manual entry page.

![screen shot 2015-01-30 at 12 11 32 pm](https://cloud.githubusercontent.com/assets/4121924/5979868/5e6a5b2e-a879-11e4-8741-b8d1bfe6a885.png)

418

kcarcia:
TESTERS: Verify that adding then deleting a location or timestamp field disables/enables the location/timestamp buttons respectively without having to refresh the page.

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

stowellm:
PASSED testing on Chrome 40.0.2214.115 (linux).

jdaigneau:
PASSED testing on Chrome 40.0.2214.115 (mac).

bdonald25:
PASSED testing on Chrome 40.0.2214.115 (linux).

Issue #1623, opened 2014-06-29
Field descriptions for edit_fields page


AlanRosenthal:
On field edit and creation page, add description for each type of field.

kcarcia:
TESTERS: Please read this and make sure they make sense.

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

stowellm:
FAILED testing on Chrome 40.0.2214.115 (linux).
For location: "'latitude, longitude' between -90 and 90." Longitude can be between -180 and 180

jdaigneau:
PASSED testing on Chrome 40.0.2214.115 (mac).

bdonald25:
PASSED testing on Chrome 40.0.2214.115 (linux).

PR #2011, opened 2015-02-19
Fields Edit_Table Redo v1

kcarcia:
Addresses #2002, #1623, #1966.

This is only the first iteration of edit_fields, therefore the buttons to add fields are at the top. That's because I haven't addressed field ordering yet, and fields are still being added through form submission. The page refreshes every time a user adds a field, and I could load the page at the bottom, but then you don't see any success or error messages. That's why, for now, the buttons are at the top.

Moving the buttons to the bottom and reordering fields will come in my second iteration of this (which you can find here #2009)

@jdalphon I do have a concern regarding a change I made to the projects controllers. The concern comes up in one of the tests. I noted where my concern is in my code.


test/functional/projects_controller_test.rb
```diff
@@ -241,8 +241,7 @@ class ProjectsControllerTest < ActionController::TestCase
                       '21_unit' => parameters['21_unit'], '22_name' => parameters['22_name'], '22_unit' => parameters['22_unit'],
                       "#{new_lat_field_id}_name" => @project.fields[num_fields - 2].name, "#{new_lat_field_id}_unit" => @project.fields[num_fields - 2].unit,
                       "#{new_long_field_id}_name" => @project.fields[num_fields - 1].name, "#{new_long_field_id}_unit" => @project.fields[num_fields - 1].unit,
-                      '20_restrictions' => 'a,b,c',
-                      '25_name' => 'Location of Foodz', '25_unit' => '', new_field: '' }, user_id: @kate.id
+                      '20_restrictions' => 'a,b,c' }, user_id: @kate.id
```
kcarcia: '25_name' => 'Location of Foodz', '25_unit' => '', new_field: '' "

This line was making the test fail. I don't understand why it's there. This test is creating that field and that line does nothing to change the field. Any thoughts?

app/controllers/projects_controller.rb
```diff
@@ -297,20 +299,23 @@ def save_fields
         redirect_to "/projects/#{@project.id}/edit_fields"
         return
       end
-    elsif field_type != ''
-      next_name = Field.get_next_name(@project,
-                                      get_field_type(params[:new_field]))
-      field = Field.new(project_id: @project.id,
-                        field_type: get_field_type(field_type), name: next_name)
+    elsif !field_type.nil?
+      while i < ((params[:num_fields].to_i > params[:text_fields].to_i) ? params[:num_fields].to_i : params[:text_fields].to_i)
```
kcarcia: Hoping implementing this won't change anything API related.

stowellm:
@kcarcia - I pulled and tested this: looks great!  As you already hinted at, it still refreshes when adding fields so it'll be awesome to see that not happen on the next iteration.

The one thing I'm not so certain about is the "Field added" pop-up when you add a field.  The first time you add a field, it shows up - great, you know you're doing it right.  The next time you add a field - assuming you haven't X'ed out the pop-up - there's now no feedback.  "Field added" just stays up there.  Possible solutions could be a timer before dismissing it, or perhaps putting the names of the fields added so that at least the text changes every time you add a field.

kcarcia:
@stowellm Great feedback, Mike. Both solutions are easy to implement. I will fix this tomorrow.

Issue #2000, opened 2015-02-12
Bug Reporting v3


kcarcia:
**General description:** Currently, there are 2 concerns regarding bug reporting:

- [x] The only place to report a bug is in the footer, and it's not easy to find. Adding the link to report a bug in other places will make it more available to users. A few of these places include <strike>(1) vis pages, (2) saved vis pages, (3) manual entry table, (4) edit_field pages,(5) projects page, (6) contact page</strike> 
- [x] Changing bug to issue.

**live/dev/localhost:** live
**iSENSE Version:** v6.0
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 40.0.2214.111

**Steps to Reproduce:** 
None

jaypoulz:
Hmmm. I suppose 'Report a Bug' falls on the negative side. In that regard, so do 'Report Problem/Issue' (assumes something bad happened).

```
Report Problem/Suggest Feature
```
Or
```
Send Us Feedback
```
Or just plain
```
Feedback
```

stowellm:
I like `Send Us Feedback`, but if I was browsing a website and found an issue, I'd immediately be using my <kbd>Ctrl</kbd>+<kbd>f</kbd> and searching for "issue" or "bug."  `Send Us Feedback` sounds like people may interpret this as sending us things like "thanks I really like this website," when really the intention is strictly to report problems.  Therefore even though the language *is* negative, it needs to be.

stowellm:
On a side note, happy 2000th issue!

kcarcia:
@stowellm @jaypoulz 

Thanks for this awesome feedback!!!

I love the idea of saying something like "Send Us Your Feedback," but I agree with Mike on both fronts: it's probably better to use "bug" or "issue" since that's likely what a user will be searching for, and I also agree "issue" is the less "harsh" word. That's the word used in MIT App Inventor, as well.

With that being said, I'll keep the alternative language in mind for other uses. For example, I think it's a great idea to add a link to the contact page. So I might say something like, "Send us your feedback here..." specifically on that page since it's likely the user shouldn't experience anything breaking on that page.

kcarcia:
Closing this as per #2094

PR #2013, opened 2015-02-23
Changing Bug to Issue, Adding More Links to Report Issues through the Site

kcarcia:
Addresses #2000


jdalphon:
Not really a fan of this. Maybe just making the link that is already in the footer more obvious would be better?

kcarcia:
Will not pull until discussed at Monday meeting.

kcarcia:
We've decided to put this in the nav bar so I'm closing this and addressing it elsewhere.

Issue #1563, opened 2014-05-23
Show pictures  on map


bdonald25:
If I upload pictures with location data I should be able to see the pictures on the map. Right now all I see is a bunch of circles showing where each picture was taken.

kcarcia:
TESTERS: Add pictures to a data set with a location and verify that picture shows up on the map at that location.

asalani93:
PASSED testing on Chrome 40.0.2214.115 x64 (Ubuntu 14.04.1 x64).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (Ubuntu 14.04 LTS).
I labeled this as pass but you may consider this a fail: I visualized two data sets, one with an image and one without.  Pictures on map aren't showing up when I click the marker(s) for the data set with the image.  They only show up when visualizing that single data set.

stowellm:
## My goats migrated from Africa to Greenland

So... I uploaded a goat to a data set with map coordinates in Africa.  When visualizing this data set, my goats show up (although left-aligned!!!).  However, when I visualize this data along with another data set that has no goats (and has points in Greenland)... my goats migrate!  Clicking Africa map points yields no goats, but clicking Greenland points suddenly yields my goats!  *And* they now are even more left-aligned with a gray box to the right.  Let's please keep my goats in Africa.  They prefer the warmer whether.  It's animal cruelty to make them swim that far anyways.

kcarcia:
TESTERS: Do your best to break this! Add pictures to a data set with a location and verify that picture shows up on the map at that location. Also verify you are viewing that data with data that does not contain pictures and preferably has other location area to catch any issue similar to Mike's migrating goats.

stowellm:
PASSED testing on Chrome 41.0.2272.101 (Ubuntu 14.04 LTS).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

PR #2015, opened 2015-02-25
View Pictures on Map Vis

bdonald25:
#1563
@jaypoulz 
I tried using a number as an indicator, but I thought that it looked much better without any type of indicator. I can put it back if you think it looks too plain.

### DIFF
--- a/app/assets/javascripts/visualizations/highvis/map.coffee
+++ b/app/assets/javascripts/visualizations/highvis/map.coffee
@@ -184,7 +184,10 @@ $ ->
               label += "<tr><td>#{field.fieldName}</td>"
               label += "<td><strong>#{dat}</strong></td></tr>"
 
-            label += "</table></div>"
+            label += '</table></div>'
+            img = data.metadata[0].photos[0]
+            if img
+              label += "<img src=#{img.src} style='width:100%'> "
 
             if groupIndex in data.groupSelection
               latlngbounds.extend latlng
--- a/app/assets/javascripts/visualizations/highvis/map.coffee
+++ b/app/assets/javascripts/visualizations/highvis/map.coffee
@@ -184,10 +184,11 @@ $ ->
               label += "<tr><td>#{field.fieldName}</td>"
               label += "<td><strong>#{dat}</strong></td></tr>"
 
-            label += '</table></div>'
             img = data.metadata[0].photos[0]
             if img
-              label += "<img src=#{img.src} style='width:90%; height: auto;'> "
+              label += "<img src=#{img.src} style='width:100%; height: auto;'> "
+
+            label += '</table></div>'
 
             if groupIndex in data.groupSelection
               latlngbounds.extend latlng
--- a/app/assets/javascripts/visualizations/highvis/map.coffee
+++ b/app/assets/javascripts/visualizations/highvis/map.coffee
@@ -193,8 +193,10 @@ $ ->
                           </div>"
 
               label +=  "</div>
-                          <a class='left carousel-control' href='#mapCarousel' role='button' data-slide='prev'><span class='glyphicon glyphicon-chevron-left'></span></a>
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/visualizations/highvis/map.coffee
```diff
@@ -173,6 +173,32 @@ $ ->
...
+                        </div>"
+
+              for i in [1...data.metadata[0].photos.length]
+                label +=  "<div class='item'>
+                            <img src=#{data.metadata[0].photos[i].src} style='width:100%; height:300px;'>
+                          </div>"
+
+              label +=  "</div>
+                          <a class='left carousel-control' href='#mapCarousel' role='button'
+                          	data-slide='prev'><span class='glyphicon glyphicon-chevron-left'></span></a>
```
jaypoulz: It's fine to use glyphicons, but would you mind trying:

```
<i class="fa fa-chevron-left"></i>
```

We don't use glyphicons anywhere else on this site so we might as well stick to font-awesome.

app/assets/stylesheets/application.css.scss
```diff
@@ -24,7 +24,7 @@ $navbar-default-color: white;
 $navbar-default-link-color: white;
 $navbar-default-link-hover-color: #DDDDDD;
 $table-bg-accent: #E8E8E8;
-
+@import "bootstrap-sprockets";
```
jaypoulz: I know this line is to fix glyphicons so leave it. Just in case we need them for something.

app/assets/javascripts/visualizations/highvis/map.coffee
```diff
@@ -184,7 +210,7 @@ $ ->
               label += "<tr><td>#{field.fieldName}</td>"
               label += "<td><strong>#{dat}</strong></td></tr>"
 
-            label += "</table></div>"
+            label += '</table></div>'
```
jaypoulz: Actually ``"`` is correct here because it's html.

HOWEVER - it's wrong to have html in js. But the vis 2.0 specs aren't written yet so maybe we'll let this slide.

app/assets/javascripts/visualizations/highvis/map.coffee
```diff
@@ -173,6 +173,32 @@ $ ->
...
+                        <div class='carousel-inner' role='listbox'>"
+
+              firstPhoto = data.metadata[0].photos[0]
+              label += "<div class='item active'>
+                          <img src=#{firstPhoto.src} style='width:100%; height:300px;'>
+                        </div>"
+
+              for i in [1...data.metadata[0].photos.length]
+                label +=  "<div class='item'>
+                            <img src=#{data.metadata[0].photos[i].src} style='width:100%; height:300px;'>
```
jaypoulz: No inline styles. Just add a class instead. There might already be one with these properties.

app/assets/javascripts/visualizations/highvis/map.coffee
bdonald25: Ok, I'll try font-awesome. Jake used them on the homepage too btw.

jaypoulz:
There is also an important thing about this pull request that was over-looked.

Any given point should only show the pictures that belong to the same data set the point belongs to.
Eg. if you are visualizing two data sets -> all points belonging to data set one should show the pictures that belong to data set 1. Ditto for data set 2.

app/assets/javascripts/visualizations/highvis/map.coffee
bdonald25: It didn't work. With font awesome the chevrons were in the top corners.

jaypoulz: y'okay. Like I said we're not really invested in either, it's just that we've been using fa icons exclusively, so now we'll have a mix. *shrug*

jaypoulz:
Two things - refresh and deal with the merge conflicts. Lastly, would you mind centering the pictures in the carousel?

bdonald25:
@jaypoulz

PR #2017, opened 2015-02-25
Adding basic kml layer support

jdalphon:
A URL to a KMLl file can be given to a project in the project/edit page. 
The map will then show that layer independent of any data that the user uploads.

There is not a test written for this yet, could @JKinsman or @bdonald25 please figure out a functional test for this, simple as setting the value and making sure its set. 

A sample kml file can be found at http://edna.usgs.gov/watersheds/sheds/USWatersheds/USWatersheds.kml


jaypoulz:
The passes my review, but let wait for travis (I've restarted the build)

Issue #2012, opened 2015-02-22
Site is getting attacked with spam user signups


fgmart:
Starting approx. Dec 5, 2014, the site has been receiving approx 30 bogus users signups per day.

E.g. see attached screenshot.

Please note not all of these users are bogus -- "Tyler P" is a developer.

The bogus logins all seem to have a last login date of Aug 16, 2013, even though they were created in the last month or two. (How is that possible.)

We need to do something about this.

![screenshot from 2015-02-22 08 58 31](https://cloud.githubusercontent.com/assets/1038856/6318767/1db4e2fc-ba71-11e4-8ca9-754b26ab525c.png)

jaypoulz:
@jdalphon - halp!

fgmart:
We talked about this and there seems to be a consensus of installing a CAPTCHA.

jdalphon:
I removed 580 accounts today that had 0 contributions.

jdalphon:
OK we are now down to 420 users!

All users that had no likes, projects, data_sets, or visualizations were axed. I backed up the database first incase anyone emails and complains.

jdalphon:
In 7 hours we had 20 new bots. Basically every 20 minutes.

kcarcia:
TESTERS: Verify the captcha on registration is working.

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

stowellm:
PASSED testing on Chrome 40.0.2214.115 (linux).

jdaigneau:
PASSED testing on Chrome 40.0.2214.115 (mac).

bdonald25:
FAILED testing on Chrome 40.0.2214.115 (linux).
For the first few users I created, the captcha was just acting like a check box and now prompting me for text. I created the users but then was unable to log back in after I logged out. Now the captcha is prompting me for text like it should and everything is working fine. I can't reproduce it.

kcarcia:
@bdonald25 Because we can't reproduce this, this will pass. However, we'll be on the lookout for this going forward.

PR #2019, opened 2015-02-26
Recaptcha

jdalphon:
Adding recaptcha so that we stop getting attacked by bots. 
We may want to upgrade to the newer version but this was the fastest solution. 
Also had to turn off some tests to get this in place quickly
#2012


jdalphon:
Failed because travis is down

config/initializers/recaptcha.rb
```diff
@@ -0,0 +1,4 @@
+Recaptcha.configure do |config|
+  config.public_key  = '6LfrsAITAAAAABT92nxUzv_-quqbsVtpu2W3uEEg'
+  config.private_key = '6LfrsAITAAAAAGF2dBUSCd-iGnX6yVpHzBNY-Yd8'
```
jaypoulz: Uhhhh. Private key's on github?

jdalphon:
I knew I was rushing too fast. Those keys are now dead.

jdalphon:
@jaypoulz  what keys...i dont see no keys ;)

PR #2020, opened 2015-02-27
Recaptcha2

jdalphon:
Adding recaptcha 2
#2012


jdalphon:
@jaypoulz @kcarcia  not sure if this should be pulled or not.

PR #2021, opened 2015-02-27
Recaptcha v2

jdalphon:
Adding recaptcha v2
Fixing problems with user creation errors

#2012

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -133,8 +133,10 @@ class UsersController < ApplicationController
     @user = User.new(user_params)
     @user.reset_validation!
 
+    captcha_message = 'Sorry, but the reCAPTCHA was incorrect.'
+
     respond_to do |format|
-      if @user.save
+      if verify_recaptcha(model: @user, message: captcha_message) && @user.save
         session[:user_id] = @user.id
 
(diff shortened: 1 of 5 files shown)


config/initializers/recaptcha.rb
```diff
@@ -0,0 +1,19 @@
+Recaptcha.configure do |config|
+
+  key_file = File.expand_path('~/.recaptcha_key').to_s
+
+  # Create the file to put Recaptcha keys in
+  unless File.exist?(key_file)
+    file = File.open(key_file, 'w')
+    file.write("REPLACE WITH PUBLIC KEY\n")
+    file.write('REPLACE WITH PRIVATE KEY')
```
kcarcia: Env variables might be best here?

app/controllers/users_controller.rb
```diff
@@ -133,8 +133,10 @@ def create
     @user = User.new(user_params)
     @user.reset_validation!
 
+    captcha_message = 'Sorry, but the reCAPTCHA was incorrect.'
```
kcarcia: Is this the error message the user sees? Hmmm.. if yes, maybe we could change it to say something like, "The text you entered in does not match the text in the image." It's possible our users don't know what recaptcha is.

Issue #1973, opened 2015-01-30
Slickgrid v2 Enhancements


kcarcia:
**General description:** The following enhancements should be made to v2 of Slickgrid:

- [x] Feedback indicating why clicking that 'X' with only a single row won't delete it (i.e. a message that says something like "Must have one row of data")

- [x] Feedback why entering an invalid lat/lon and click 'add row' erases data.

- [x] If a user enters invalid data and then clicks save, the data is simply wiped out. We probably should disable save if input is invalid or show error messages on save or some variation of this.

- [x] More specific title for /manualEntry /edit. Current title is "Data_sets"

- [x] Add Cancel button which takes you back to the project page.

- [x] Default to 10 rows of data.

**live/dev/localhost:** live
**iSENSE Version:** v6.0
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 40.0.2214.94

**Steps to Reproduce:** 
1. Go to a project.
2. Click manual entry.

asalani93:
In regards to the third bullet point, this is currently not possible with the current way I have the code structured.  There is either no way to force the grid to valid its data, or the documentation is lacking even more than I thought it was.

However, it is important to fix that.  Therefore, I've decided to go back and refactor/redo a lot of the slickgrid stuff.  The other three are super easy fixes, though.  This should be done before the next release.

jaypoulz:
@asalani93 is it not possible to have a backup copy of the data made on save, and repopulate the grid with the backup upon failure?

asalani93:
There's no way to do that without either allowing invalid data to leak into
the dataset, or killing compatibility with any non-in-house editors.

jaypoulz:
@asalani93 so slickgrid has limited export / import capabilty? I mean, export has to be there for uploading to the website, but import isnt?

asalani93:
Importing/exporting isn't the problem - that's incredibly easy.  The problem is that when saving, I currently force the current edit to try and commit.  If it's an invalid edit, it's discarded.  This is mostly all default Slickgrid behavior

Repopulating the grid would put the data as the initial loaded value for the editor, which is implicitly accepted as valid.  You could put an invalid latitude and longitude into a dataset, click save, the editor remembers the current edit, it fails validation, it puts back the invalid edit and opens the editor.

[...]

PR #2029, opened 2015-03-03
Slickgrid V2 enhancements

asalani93:
Addresses #1973, #2044.

* Cell in first row/column opens automatically
* I added a button to open the datetimepicker instead of opening it when you select a timestamp cell
* There are ten rows by default
* Holding down enter gives you more row
* Better title for manual entry
* More popovers when users do dumb things
* Fixed field validation.  Now clicking add row with bad data doesn't nuke the bad data.  Only downside is that now you can't delete rows if you have bad data currently open.  This is a slickgrid problem.
* Fixed broken tests
* Added a gem that lets you screenshot the page during integration tests
* Minor stylistic code changes
* Fixed button colors
* Implements wrapping for left and right cell movement
* Units in the column headers
* Subsequent validation errors are placed correctly
* Text fields with restrictions no longer overflow if the column is too tiny


kcarcia:
@asalani93 A few things,

- [ ] Cancel button needs to be red. btn-danger is the correct class.
- [ ] Add Row and Save colors should be switched.
- [ ] Allowing tab when you're at the end of the form would also be great. Is this possible?

Otherwise, really nice job.

app/assets/javascripts/data_sets/editTable.js.coffee
```diff
@@ -172,60 +263,33 @@ showError = (error) ->
...
   data = $('#slickgrid-container').data 'data'
-
-  rets = setupTable cols, data
-  grid = rets[0]
-  view = rets[1]
-
-  setupResizer grid
+  grid = new Grid cols, data,
+    url: "#{uploadSettings["urlEdit"]}"
+    type: "#{uploadSettings["methodEdit"]}"
```
kcarcia: " nested in " quotes?

app/assets/javascripts/data_sets/editTable.js.coffee
```diff
@@ -17,148 +17,239 @@ uploadSettings =
...
-    grid.invalidateRow args.rows
-    grid.render()
-
-  $('.edit_table_add').click ->
-    if $('.edit_table_save').hasClass 'disabled'
-      return
-
-    grid.getEditorLock().commitCurrentEdit()
-    grid.resetActiveCell()
+Grid = (cols, data, submit) ->
```
jaypoulz: this should be camelcase, so ```grid```

app/assets/javascripts/data_sets/editTable.js.coffee
```diff
@@ -17,148 +17,239 @@ uploadSettings =
...
+      autoEdit: true
+      editable: true
+      enableCellNavigation: true
+      enableColumnReorder: false
+      forceFitColumns: true
+      rowHeight: 35
+      syncColumnCellResize: true
+
+    # associate IDs with each row in the table
+    for _, i in data
```
jaypoulz: Interesting use of the underscore character

app/assets/javascripts/data_sets/editTable.js.coffee
```diff
@@ -17,148 +17,239 @@ uploadSettings =
...
+
+    grid.onKeyDown.subscribe (e, args) ->
+      if e.keyCode == 13 and grid.getDataLength() - 1 == grid.getActiveCell().row
+        addRow()
+
+    $(window).trigger 'resize'
+
+    setTimeout ->
+      $('.slick-cell.l0.r0').first().trigger 'click'
+    , 1
```
jaypoulz: is there a reason this can't be on the line above?

Also, aside from function definitions, the optional parens can help to group arguments neatly

vendor/assets/javascripts/slickgrid-editors/slickgrid-time.coffee
```diff
@@ -5,61 +5,79 @@
 
 @TimestampEditor = (args) ->
   form = null
-  loadValue = null
   currValue = null
-  canClose = false
+  loadValue = null
+  pickerOpen = false
+
+  form = $('<div>
```
jaypoulz: can this form not be given an 'id' to query upon? this is grosssss

jaypoulz:
Overally, nice patch! A few things are a little weird, but then again so is js.

vendor/assets/javascripts/slickgrid-editors/slickgrid-time.coffee
asalani93: It can, but this is better.  Querying is slow.  We must be fast.

app/assets/javascripts/data_sets/editTable.js.coffee
asalani93: I maintain that it should be capital, but I will change it because I have picked my battles and this is not one of them.

app/assets/javascripts/data_sets/editTable.js.coffee
asalani93: Do I change it?  This pretty standard in lots of functional languages when you want to ignore a parameter.

asalani93:
Ready for review again:
* Minor stylistic code changes
* Fixed button colors
* Implements wrapping for left and right cell movement
* Units in the column headers

asalani93:
The test this is failing is caused by an actual bug - deleting a project doesn't remove the project.  I can guarantee this is not caused by my pull request.

kcarcia:
@asalani93 no worries, we know. I fixed a broken test, and now it's failing sporadically. Jeremy is working to fix it

jaypoulz:
Turns out the problem is really, REALLY bad ajax. As soon as I figure out what is going on at a deeper level, I'll submit a patch

app/assets/javascripts/data_sets/editTable.js.coffee
jaypoulz: Nah it's cool, I've just never seen it used that way

kcarcia:
@asalani93 

2 last small changes,
- [ ] As noted above, please convert the " quotes into ' quotes  in your code as we discussed.
- [ ] Found one bug. Basically, if you get 2 validation errors on two different fields on the manual entry form, the second validation message appears where the first one did. To reproduce, (1) go to a project with 3 number fields; (2) enter invalid data into the first number field, note the validation message; (3) enter valid data into the second number field; (4) enter invalid data once again in the third number field, and notice the validation message appears on the first number field (where you the first validation error message appeared).

Overall, great patch. Please make these last 2 changes and assign back to me.

kcarcia:
@asalani93 What was Tyler's bug? Just want to make sure I test that tomorrow when I review.

asalani93:
Tyler's bug was that text fields with restrictions would overflow their cell and go into the next one if the columns were too tiny.

Issue #1897, opened 2014-12-19
Histogram as a stacked bar of individual points


fgmart:
**General description:**
For pedagogical reasons, histogram view should be enhanced so that each bar is represented as a stack of individual data items.

This will help people realize what the histogram actually means: for each bin, a count of data elements that fit into that bin.

I realize that some logic will be required to determine how/whether to draw the bar as individual elements; e.g., if there is a bar that's 500px high which represents 1000 elements, the feature should automatically disable.

But assuming it can reasonably be drawn, this would be a big improvement to histogram.

**live/dev/localhost:** N/a
**iSENSE Version:** N/a
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** N/a
**Browser/Version:** N/a

**Steps to Reproduce:** N/a

**Associated Image(s):**
![12_01](https://cloud.githubusercontent.com/assets/1038856/5507909/8fabbe0a-8775-11e4-8f1d-faa3e3953ed7.jpg)

fgmart:
Hey all,

This is something I would like to put on the short list. 

Is there a chance that @stowellm can work on this?

Thanks.

jaypoulz:
@fgmart I think @JKinsman would be more comfortable making this type of change for now. Assuming we haven't already added too much to his queue. :smile: Otherwise, we might have to have @stowellm test his mettle against this patch.

JKinsman:
@fgmart @jaypoulz I should have a chance to look into this.  Assuming the regression patch gets merged (once I make the styling change requested), I could do it once I refactor the homepage, which I should be able to finish tonight.

kcarcia:
TESTERS: Try to break historgram!

asalani93:
PASSED testing on Chrome 40.0.2214.115 x64 (Ubuntu 14.04.1 x64).
maybe give each item in the histogram its actual value instead of the column value

stowellm:
PASSED testing on Chrome 41.0.2272.89 (Ubuntu 14.04 LTS).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

PR #2030, opened 2015-03-04
Histogram as a stacked bar of individual points

JKinsman:
Addresses issue #1897 .  Assuming the amount of data visualized on the histogram is not unreasonable, it will behave as requested.

The feature will automatically disable if:
1.  Any series has more than fifty points in a single bin.  This metric was chosen because of limitations to the Highcharts API.  A highcharts series cannot have multiple data points with the same x-axis value.  Large bins from individual groups result in adding a considerable number of series to the chart, and quickly slow down performance.
2.  Any bin has more than one hundred points between all groups that are currently selected.  For similar reasons, bins with a large number of data points quickly degrade performance.  A threshold of 100 points was chosen because it seems to be around where older machines begin to slow down.

@jaypoulz


app/assets/javascripts/visualizations/highvis/histogram.coffee
```diff
@@ -198,30 +198,63 @@ $ ->
...
+              # Get total for this bin
+              for dc, groupData of binObjs
+                if groupData[number]
+                  sum += groupData[number]
+
+              ret =
+                x: (Number number)
+                y: occurences
+                total: sum
+          ## --- ###
```
jaypoulz: this comment can probably go :smile: I don't think It's been useful for a while haha.

app/assets/javascripts/visualizations/highvis/histogram.coffee
```diff
@@ -198,30 +198,63 @@ $ ->
...
+                color: globals.configs.colors[Number(group) % globals.configs.colors.length],
+                name: data.groups[Number(group)]
+                data: binData
+              @chart.addSeries options, false
+              i += 1
+            i = 0
+        else
+          for groupIndex in data.groupSelection
+            finalData = for number, occurences of binObjs[groupIndex]
+
```
jaypoulz: no newline here

JKinsman:
@jaypoulz should be good to go once the newest commit passes Travis! :smile:

jaypoulz:
I'm still opposed to this, but the patch is so nice  I can't help but r+.

One small note though - in the future, please keep lines < 80 characters.

PR #2033, opened 2015-03-04
Added report issue link to nav bar

kcarcia:
screen shot 2015-03-04 at 11 49 42 am

@jdalphon @fgmart @jaypoulz Is this what we're going for?

Addresses #2000.

### DIFF
--- a/app/views/home/create_issue.html.erb
+++ b/app/views/home/create_issue.html.erb
@@ -2,7 +2,7 @@
   <div class="panel panel-default">
     <div class="panel-heading">
       <div class="row">
-        <div class="col-md-9 h1">Report a Bug</div>
+        <div class="col-md-9 h1">Report Issue</div>
       </div>
     </div>
     <div class="col-md-12">
--- a/app/views/home/report_bug.html.erb
+++ b/app/views/home/report_bug.html.erb
@@ -1,10 +1,10 @@
 <div class="panel panel-default">
   <div class="panel-heading">
-    <h1>Report a Bug: Do you have a GitHub account?</h1>
+    <h1>Report Issue: Do you have a GitHub account?</h1>
   </div>
 
   <div class="panel-body">
-    <p>All bug reports will be posted to our public issue repository on GitHub. Don't have a GitHub account? No problem! You can still report bugs.</p>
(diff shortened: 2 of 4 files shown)


jaypoulz:
@kcarcia I think we're dropping the "News" label in favor of featuring it (News articles) on the new homepage.  My personal opinion is that it should go where register is when register isn't there.

kcarcia:
@jaypoulz @jdalphon @fgmart 

To have it in place of the  News tab is fine. The News tab will have to present for admins still, though. Unless we redo how admins add tutorials.

As far it replacing the register, I'm not quite sure how that'll work. When users are logged in, the register and login buttons become a button to their account and the logout button. I don't think we'll be able to replace either of those. But would you prefer to add it to the right set of buttons?

jdalphon:
We have already had complaints that you have to log out to register a new
user (when teachers are batch creating accounts). Although that button goes
away for every website ever. I did see a website last night that had better
wording that people ''''''might''''''' like and which wouldnt have to be in
the header.

How are we doing? Give us feedback on this page.

Appears on several of the pages on rei.com

On Thu, Mar 5, 2015 at 11:14 AM, Kaitlyn Carcia <notifications@github.com>
wrote:

> @jaypoulz <https://github.com/jaypoulz> @jdalphon
> <https://github.com/jdalphon> @fgmart <https://github.com/fgmart>
>
> To have it in place of the News tab is fine. The News tab will have to
> present for admins still, though. Unless we redo how admins add tutorials.
>
> As far it replacing the register, I'm not quite sure how that'll work.
> When users are logged in, the register and login buttons become a button to
> their account and the logout button. I don't think we'll be able to replace
> either of those. But would you prefer to add it to the right set of buttons?
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/isenseDev/rSENSE/pull/2033#issuecomment-77393565>.
>



-- 
--------------------------------
James Dalphond
UML Computer Science
Engaging Computing Group

kcarcia:
@jaypoulz @jdalphon @fgmart 

I like Jim's idea, but it's not clear to me how to implement it. Would this statement be on every day? If not, what pages? Where would it appear? Do we all like this idea?

kcarcia:
Closing as per #2091 since the report bug option is no longer hidden.

Issue #2032, opened 2015-03-04
Reorder tabs on user page


kcarcia:
**General description:** The order of the tabs on the user page should be as follows:
My Projects
Data Sets
Visualizations
Liked Projects

**live/dev/localhost:** live
**iSENSE Version:** v6.3
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 40.0.2214.115

**Steps to Reproduce:**
1. Login.
2. Go user page.

kcarcia:
TESTERS: The Liked pane should now be the last tab on the user's page.

asalani93:
PASSED testing on Chrome 40.0.2214.115 x64 (Ubuntu 14.04.1 x64).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (Ubuntu 14.04 LTS).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

PR #2034, opened 2015-03-04
Switched tab order

kcarcia:
Addreses #2032


jaypoulz:
The test failure is a bit weird, but the patch looks solid.

kcarcia:
@jaypoulz Yes, it is. However, there was no previous place on that page that said "Delete this project"

Issue #2036, opened 2015-03-06
Can't template fields with a CSV that has only headers


stowellm:
**General description:** I tried setting up fields for a project using a CSV that had only headers.  Got an error (see image). However if I add even one number to the row below the headers, all is fine.

**live/dev/localhost:** dev
**iSENSE Version:** v6.0-316-g896eb33
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu LTS 14.04
**Browser/Version:** Chrome 40

**Steps to Reproduce:**
1. Create new project
2. Setup fields by uploading file
3. Choose a CSV with only headers (no data)
4. Observe error as seen in the image below

**Associated Image(s):**
![something](https://cloud.githubusercontent.com/assets/1039162/6527177/bb508c40-c3e6-11e4-9821-5281ac834113.png)

jdalphon:
Its trying to figure out what the types  are and needs to see data

On Fri, Mar 6, 2015 at 9:54 AM, Mike Stowell <notifications@github.com>
wrote:

> *General description:* I tried setting up fields for a project using a
> CSV that had only headers. Got an error (see image). However if I add even
> one number to the row below the errors, all is fine.
>
> *live/dev/localhost:* dev
> *iSENSE Version:* v6.0-316-g896eb33
> *Logged in (Y or N):* Y
> *Admin (Y or N):* N
>
> *OS:* Ubuntu LTS 14.04
> *Browser/Version:* Chrome 40
>
> *Steps to Reproduce:*
> 1. Create new project
> 2. Setup fields by uploading file
> 3. Choose a CSV with only headers (no data)
> 4. Observe error as seen in the image below
>
> *Associated Image(s):*
> [image: something]
> <https://cloud.githubusercontent.com/assets/1039162/6527177/bb508c40-c3e6-11e4-9821-5281ac834113.png>
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/isenseDev/rSENSE/issues/2036>.
>



-- 
--------------------------------
James Dalphond
UML Computer Science
Engaging Computing Group

stowellm:
Yup - I believe I just fixed this.  It need data only for timestamp/text fields, so I switched the ordering to check for latitude/longitude first (since this only requires the name of the field), then checks if the data column is nil.  If so, it continues the loop rather than failing.

kcarcia:
[...]

PR #2039, opened 2015-03-06
Can template fields using a csv with only headers, no data

stowellm:
Addresses #2036 :cactus: 

I switched the field matching around to first check for latitude/longitude before timestamp/text.

Latitude and longitude requires checking `column[0]` content - the header of the CSV - to perform field matching.  The other field matching checks require `column[1]` - the data of the CSV.

Therefore the logic has been changed as follows: 
- check lat/lon
- check if data in column[1]
-  if no data, continue to next field, otherwise continue checking for timestamp/text


app/views/projects/show.html.erb
```diff
@@ -185,8 +185,16 @@
       <%= render layout: "shared/box", locals: {title: 'Fields', width: '12', unique_id: 'fields', should_hide: false} do %>
         <% if (@project.fields.count == 0) && can_edit?(@project) %>
           <div class="col-md-12 center">
-            <a href="/projects/<%=@project.id%>/edit_fields"><button class="btn btn-success" id="manual_fields"><i class="fa fa-keyboard-o" style="margin:5px"></i>Setup Manually</button></a>
-            <button class="btn btn-success" id="template_file_upload"><i class="fa fa-file-o" style="margin:5px"></i>Upload File</button>
+            <a href="/projects/<%=@project.id%>/edit_fields">
```
jaypoulz: spaces -> ``<%= project.id %>``

app/views/projects/show.html.erb
```diff
@@ -185,8 +185,16 @@
...
+              <button class="btn btn-success" id="manual_fields">
+                <i class="fa fa-keyboard-o" style="margin:5px"></i>
```
jaypoulz: having a bunch of icons with the same inline style is wat. either have a class for this, or add the style to i tags in general if it's universal

lib/file_uploader.rb
```diff
@@ -125,6 +125,10 @@ def swap_columns(data_obj, project)
   end
 
   def swap_with_field_names(data_obj, project)
```
jaypoulz: Add a comment describing wat is data_obj

lib/file_uploader.rb
```diff
@@ -125,6 +125,10 @@ def swap_columns(data_obj, project)
...
+    if data_obj.first[1].nil?
```
jaypoulz: wat is data_object.first, and in turn, wat is data_obj.first[1]

lib/file_uploader.rb
```diff
@@ -125,6 +125,10 @@ def swap_columns(data_obj, project)
...
+      return []
+    end
+
     data = []
```
jaypoulz: return data, by defining earlier

lib/file_uploader.rb
```diff
@@ -203,14 +207,18 @@ def get_probable_types(data_set)
...
 
     data_set.each do |column|
+      if column[0].casecmp('LATITUDE') == 0 or column[0].casecmp('LAT') == 0
+        types['latitude'].push column[0]
+        next
+      elsif column[0].casecmp('LONGITUDE') == 0 or column[0].casecmp('LON') == 0
+        types['longitude'].push column[0]
+        next
+      end
+      next if column[1].nil?
```
jaypoulz: and a comment saying this is the no data case

app/views/projects/show.html.erb
stowellm: It's only these two.  Out of every single `<i>` tag on rSENSE, it's only this guy with `fa-keyboard-o` and `fa-file-o`.  And those classes don't even live in /rSENSE (must come from somewhere else)?  Do you *really* want another class in the CSS just to add a `margin:5px` to two icons?

lib/file_uploader.rb
stowellm: Done - diff not being suppressed on Github because comment is one line below.

app/views/projects/show.html.erb
jaypoulz: Do I want a class, kindof. :smile: But we can let it go if it's two icons I guess. Just seems silly to have inline styles when you can declare page specific styles, and then your style doesn't have to bleed into the html.

stowellm:
@jaypoulz ready for re-review

jaypoulz:
Interesting side effect of this patch - I just uploaded a csv with no data, and the fields were matched nicely, but it also made an empty data set for me.

stowellm:
That is correct behavior.  If you choose to save a data set but happen to have no data, behold: you have no data.

jaypoulz:
The default could be better. :unamused:

stowellm:
Fine... but you're a :cactus: :smile:

jaypoulz:
I understand why you did you what you did, but I really doubt it's necessary. The data_obj variable should be available at that point in the code, assuming the projects controller is the one rendering this page. Just because it was done with a class property in one part of the code does not necessarily deem it necessary to do in another part of the code. That being said, I'm just going to check out your branch and make the change myself after my exam tonight, because I want this to get through into the release tomorrow.

stowellm:
Y'okay. I'll check out what you did tomorrow morning to see what you wanted to change, as I'm still a little confused.

jaypoulz:
Turns out what I wanted wasn't possible. I didn't realize the source of what gets passed into data_obj was from another rails method. There is a CoffeeScript solution to do what I want, but it's arguably worse than this implementation.

kcarcia:
:squirrel:

Issue #2006, opened 2015-02-13
API Documentation Copyedits


kcarcia:
**General description:**
- [x] /api/v1/projects/XX --- "•XX should be replaced with the ID for the project. Or set id as a GET variable" --- incoherent
- [x] /api/v1/fields/XX --- same as above
- [x] /api/v1/projects --- "Response: Projects (array)" --- JSON format this please
- [x] /api/v1/fields --- "•restrictions - Currently text fields only. Defines accepted values "blue,green,red" will force the project to accept only those values" --- incoherent
- [x] /api/v1/data_sets/append --- put "none" or "n/a" or something instead of blank space under "Notes"
- [x] /api/v1/users/myInfo --- same as above
- [x] /api/v1/media_objects --- "•type - either "project", or "data_sets" --- cringe-worthy grammar

**live/dev/localhost:** live
**iSENSE Version:** v6.0
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 40.0.2214.111

**Steps to Reproduce:** None.

1

jaypoulz:
TESTERS: Once again, only check this out if you are working on/with the API. Please be on the lookout for any spelling mistakes or inconsistencies.

stowellm:
The build thingy in the top right run off the page.  Can it just be pushed left like 20px?

![build_thing](https://cloud.githubusercontent.com/assets/1039162/6872276/9a08eb06-d47e-11e4-9656-ace290263874.png)

stowellm:
FAILED testing on Chrome 41.0.2272.101 (Ubuntu 14.04 LTS).
`Example (HTML) **originial** code by Scott Cytacki`

Again, still worth pulling.  Make a new issue for this though.

Issue #1987, opened 2015-02-10
API Extra Project calls not listed


tylerpuleo:
General description: The API has extra calls to projects that aren't listed in the Docs. "/api/v1/projects"

live/dev/localhost: live
iSENSE Version: v6.0
Logged in (Y or N): Y
Admin (Y or N): N

OS: Ubuntu 14.04 (Posting with Javascript)
Browser/Version: Chrome 41.0.2272.43

Steps to Reproduce: N/A

jaypoulz:
I think what is meant here is that the are optional parameters to projects -> index that are not listed in the documentation

tylerpuleo:
Yes, that's a much better way of explaining it thank you Jeremy.

jaypoulz:
TESTERS: There is a lot of new documentation here. If you work a lot with the API, I'd appreciate it if you could glance over the documentation and try to find typographical errors or inconsistencies. If we find any, we'll follow this up with a cleanup patch.

JKinsman:
FAILED testing on Safari 8.0.3 (Mac OSX 10.10.2).
Documentation looks good, although the information that appears when an API call is clicked overflows the main page.

PR #2042, opened 2015-03-06
Better API documentation

bdonald25:
#2006 #1987
@jaypoulz This is not ready to pull until I get some examples for the media object api calls. However, it would be great to get some feedback on the rest of the page.


jaypoulz:
Here is my feedback:

- [ ] you know that build failing image at the top of the page? Trying changing it's source to:
``src="https://travis-ci.org/isenseDev/rSENSE.svg?branch=<%= @version %>"``
This will be a broken link for your local host and rsense-dev, but should check against the latest stable build in live, reducing the odds that you'll see "build failing" on the page.
- [ ] "Ok" should either by "OK" or "okay". I prefer the former.
- [ ] Whenever you list an object (link in an ajax call) in the documentation, line up the colons
- [ ] You know how you have XX to represent a variable in the doc line? Why not use that same variable in the code to represent what is missing. You can leave it bold, but using the same variable name would give a sense of internal consistentency
- [ ] The get projects docs are all kinds of wrong
- [ ] It's not obvious how to expand the docs (I mean, who reads directions!). It might be cool if they simply expanded on hover over, or maybe if there was a highlight effect on hover. Maybe the whole header part should be clickable instead of just the request type.

kcarcia:
@bdonald25 It looks like you are ready for review again since you added that tag so I reassigned @jaypoulz

jaypoulz:
I made a pull request to your branch, if that merges and passes testing, it'll be done.

jaypoulz:
CAUTION: DO NOT PULL UNTIL #2077 HAS LANDED

Issue #2025, opened 2015-02-27
Deleting an image from a data set redirects to project page


stowellm:
**General description:** When visualizing a single data set with media, deleting the media redirects you back to the project's show page instead of the data set visualization.

**live/dev/localhost:** dev
**iSENSE Version:** v6.0-156-gb3d8fba
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu 14.04 LTS
**Browser/Version:** Chrome 40.0

**Steps to Reproduce:**
1. Visualize a **single** data set
2. Add media
3. Remove the media
4. Note you aren't redirected back to the visualization you were just on, but rather the project page

**Associated Image(s):**

kcarcia:
TESTERS: Follow the steps to reproduce to test this.

asalani93:
PASSED testing on Chrome 40.0.2214.115 x64 (Ubuntu 14.04.1 x64).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (Ubuntu 14.04 LTS).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

PR #2049, opened 2015-03-12
Better redirect on media delete

stowellm:
Addresses #2025 

Rather than redirecting to the media_object's parent (always the project page), we now redirect back to wherever we were before.  Hooray. :camel:


asalani93:
this is good

stowellm:
Had to fix something (test was breaking because the redirect path was not set) - @asalani93 mind rereviewing?

Issue #2053, opened 2015-03-12
Can't edit data set name


fgmart:
**General description:** I can't rename a data set. It flashes red when I try to save.  
Also, from the Edit Data interface, it makes me think that it's working (no red flashes), but when I save the whole thing, it doesn't actually rename.

**live/dev/localhost:** live
**iSENSE Version:** v6.3
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 40.0.2214.111

**Steps to Reproduce:** Doing the above.

12

jaypoulz:
Awesome - that's for using the report a bug button! :smile: 

Could we possibly get the project + data set you were working on as well?

kcarcia:
@fgmart @jaypoulz Looks like this is actually happening across the board.

jaypoulz:
Excellent - let's up it to high priority and let @stowellm loose on it!

jdalphon:
We were setting the url for the ajax request, then throwing it away.
``` javascript
href = location.pathname
```
fix
```javascript
 href = if ($ @).attr('href')? then ($ @).attr('href') else location.pathname
```

Testing now

kcarcia:
TESTERS: Verify editing a data set name works now.

asalani93:
PASSED testing on Chrome 40.0.2214.115 x64 (Ubuntu 14.04.1 x64).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (linux).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (linux).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (linux).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (linux).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (Ubuntu 14.04 LTS).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (Ubuntu 14.04 LTS).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (Ubuntu 14.04 LTS).

stowellm:
PASSED testing on Chrome 41.0.2272.89 (Ubuntu 14.04 LTS).

stowellm:
[...]

PR #2055, opened 2015-03-13
Fixing data set rename error.

jdalphon:
Towards issue #2053

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -258,3 +258,12 @@ body {
     }
   }
 }
+
+.half_width_edit
+{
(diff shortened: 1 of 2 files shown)


stowellm:
Fails my review.  Reason: log in, create a project with a number field, add two data sets, visualize both, then log out.  You get a flash "Logged Out" then get redirected *again* with an error flash "Unauthorized".  

From here, you can hit back on your web browser: but it brings you back to the time it said "Logged Out."  You have to hit back *again* to get back to where you were before this visualization.

stowellm:
Passes my review.  I had originally left a comment regarding another issue I discovered but it's not due to this pull request.  I'll open a new issue for that.

Issue #2062, opened 2015-03-13
Default type for empty CSVs should be number, not text


asalani93:
**General description:** Currently, empty CSV fields default to number, not text.  Text is more general.

**live/dev/localhost:** dev
**iSENSE Version:** v6.3
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu
**Browser/Version:** Chrome

**Steps to Reproduce:**
Upload an CSV with no fields.  Observe as they default to number.

**Associated Image(s):**
:dog:

stowellm:
We decided number is a better default.  iSENSE likes numbers.  Nevertheless, I improved the field matching for CSVs.  See the PR for details.

jaypoulz:
Yesh. Text might be more generic than numbers, but numbers are more useful than text when generating visualizations. As much as possible, our goal is to encourage users to generate meaningful vises. As a general concept, the barrier between the two types will erode as we introduced numbers as grouping fields, and text as enums.

kcarcia:
TESTERS: Upload a CSV with only headers. Every field should match to a number, except a column with the header "Time" (in that case, it should match to timestamp)

stowellm:
PASSED testing on Chrome 41.0.2272.101 (Ubuntu 14.04 LTS).

stowellm:
PASSED testing on Chrome 41.0.2272.101 (Ubuntu 14.04 LTS).

JKinsman:
PASSED testing on Safari 8.0.3 (Mac OSX 10.10.2).

PR #2072, opened 2015-03-23
CSV templates default to *number* if data are nil

stowellm:
Addresses #2062 

Additionally, will match to Timestamp if the data is nil but the header contains 'time'

### DIFF
--- a/lib/file_uploader.rb
+++ b/lib/file_uploader.rb
@@ -219,7 +219,7 @@ class FileUploader
       # is nil but the header contains "time," classify as a timestamp
       if (!column[1].nil? and
            (column[1]).map { |dp| (regex =~ dp) }.reduce(:&)) or
-           (column[1].nil? and column[0].upcase().include?('TIME'))
+           (column[1].nil? and column[0].upcase.include?('TIME'))
         types['timestamp'].push column[0]
         next
       end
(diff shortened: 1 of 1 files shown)


lib/file_uploader.rb
```diff
@@ -215,19 +215,30 @@ def get_probable_types(data_set)
...
+      if column[0].casecmp('LONGITUDE') == 0 or column[0].casecmp('LON') == 0
         types['longitude'].push column[0]
         next
       end
-      next if column[1].nil? # no data case
-      if (column[1]).map { |dp| (regex =~ dp) }.reduce(:&)
-        types['timestamp'].push column[0]
-      elsif !(column[1]).map { |dp| valid_float?(dp) }.reduce(:&)
+      # if there is either no data or the data is not a valid float,
+      # classify as text
```
jaypoulz: this comment is wrong

lib/file_uploader.rb
```diff
@@ -215,19 +215,30 @@ def get_probable_types(data_set)
...
+      if !column[1].nil? &&
+          !(column[1]).map { |dp| valid_float?(dp) }.reduce(:&)
```
jaypoulz: this indentation is different from the other multiline ifs in this file

Issue #2045, opened 2015-03-07
Contributor key icon is disproportionately scaled


fgmart:
The contributor key looks bad because it's not proportionally scaled. Here's how it looks on the web site:

![contrib-key-squashed](https://cloud.githubusercontent.com/assets/1038856/6543542/8f8651dc-c4ed-11e4-9714-088ceff0284a.png)

Here's the actual image:

![key](https://cloud.githubusercontent.com/assets/1038856/6543545/a4994692-c4ed-11e4-8607-b371a656d6bd.png)

The key shown should be proportionally scaled so it's not stretched vertically.

kcarcia:
Also, the Submit key button and key input box need some margin. https://cloud.githubusercontent.com/assets/1039162/6802342/f77185da-d205-11e4-8521-b9d93ea11dc6.png

jaypoulz:
TESTERS: Verify that the "Submit Key" button has some margin, and that the contributor key image is no longer being stretched.

stowellm:
PASSED testing on Chrome 41.0.2272.101 (Ubuntu 14.04 LTS).

PR #2075, opened 2015-03-23
Resized key image

kcarcia:
Do you think it looks better?

Addresses #2045


stowellm:
That does look much better; no more squished key!  

Something else that I did notice when entering a key to contribute to a project is that the "Submit Key" button and "Key" text field have no layout margin between them (i.e. they're too close together); do you agree?

button_too_close

kcarcia:
@stowellm 100% I'll modify the bug and fix that tomorrow. :)

kcarcia:
@stowellm Fixed that, Mike. It's ready for review again.

stowellm:
@kcarcia Looks great!  Passes my review.

Issue #1988, opened 2015-02-10
API has no error for making contrib key if key already exists


tylerpuleo:
General description: If contrib key already exists and you try to make another contrib key with the same name there is no error message thrown "/api/v1/projects/XX/add_key"

live/dev/localhost: live
iSENSE Version: v6.0
Logged in (Y or N): Y
Admin (Y or N): N

OS: Ubuntu 14.04 (Javascript)
Browser/Version: Chrome 41.0.2272.43

Steps to Reproduce: Run this Javascript:

    var apiUrl = 'http://isenseproject.org/api/v1/projects/835/add_key';

    var upload = {

        'email': 't@t.t',
        'password': 't',
        'contrib_key': {'name' : 'key','project_id' : 835, 'key' : 'key'}
    }
    $.post(apiUrl, upload);

jaypoulz:
TESTERS: You can go ahead and ignore this issue. @tylerpuleo Would you mind trying out a few of your JS requests against the API and verifying this?

tylerpuleo:
Absolutely

tylerpuleo:
Bug no longer a bug

PR #2077, opened 2015-03-24
Report error for trying to create a key through API if it exists

bdonald25:
#1988

### DIFF
--- a/app/controllers/api/v1/projects_controller.rb
+++ b/app/controllers/api/v1/projects_controller.rb
@@ -67,42 +67,53 @@ module Api
           respond_to do |format|
             format.json { render json: { msg: 'contrib_key is nil' }, status: :unprocessable_entity }
           end
-        elsif params[:contrib_key][:name].nil?
+        return
+        end
+
+        if params[:contrib_key][:name].nil?
           respond_to do |format|
             format.json { render json: { msg: 'name is nil' }, status: :unprocessable_entity }
           end
-        elsif params[:id].nil?
+        return
+        end
+
+        if params[:id].nil?
           respond_to do |format|
             format.json { render json: { msg: 'project id is nil' }, status: :unprocessable_entity }
           end
-        elsif params[:contrib_key][:key].nil?
+        return
+        end
+
+        if params[:contrib_key][:key].nil?
           respond_to do |format|
             format.json { render json: { msg: 'key is nil' }, status: :unprocessable_entity }
           end
-        else
-          @project = Project.find(params[:id])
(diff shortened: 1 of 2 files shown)


jaypoulz:
It looks like your failing your own test :/

bdonald25:
Yeah, so my tests don't work locally. I have not been able to get them working correctly since my laptop was wiped so I have to let travis test them for me, just until I can figure out what is wrong.

jaypoulz:
```
def add_key
        if params[:contrib_key].nil?
          respond_to do |format|
            format.json { render json: { msg: 'contrib_key is nil' }, status: :unprocessable_entity }
          end
        return
        end
        
        if params[:contrib_key][:name].nil?
          respond_to do |format|
            format.json { render json: { msg: 'name is nil' }, status: :unprocessable_entity }
          end
        return
        end
        
        if params[:id].nil?
          respond_to do |format|
            format.json { render json: { msg: 'project id is nil' }, status: :unprocessable_entity }
          end
          return
        end

        if params[:contrib_key][:key].nil?
          respond_to do |format|
            format.json { render json: { msg: 'key is nil' }, status: :unprocessable_entity }
          end
          return
        end

          @project = Project.find(params[:id])
          unless @cur_user.id == @project.user_id
           respond_to do |format|
              format.json { render json: { msg: 'User does not own the project.' }, status: :unauthorized }
            end
           return
           end

            key = @project.contrib_keys.find_by_key(params[:contrib_key][:key])
            unless key.nil?
              respond_to do |format|
                format.json { render json: { msg: 'key already exists.' }, status: :conflict }
              end
              return
            end

             key = ContribKey.new(name: params[:contrib_key][:name], project_id: @project.id, key: params[:contrib_key][:key])
              respond_to do |format|
                if key.save
                  format.json { render json: { msg: 'Success' }, status: :created }
                else
                  format.json { render json: { msg: 'Unprocessable Entity.' }, status: :unprocessable_entity }
                end
              end
          end
        end
      end
    end
  end
end
```

app/controllers/api/v1/projects_controller.rb
```diff
@@ -67,43 +67,55 @@ def add_key
           respond_to do |format|
             format.json { render json: { msg: 'contrib_key is nil' }, status: :unprocessable_entity }
           end
-        elsif params[:contrib_key][:name].nil?
+        return
```
jaypoulz: these returns should line up with the respond_to block

jaypoulz:
Sirens! :cop:

PR #2081, opened 2015-03-27
Added a get latest data to the regular vis page

jaypoulz:
No associated issue, but implements the refresh feature that is currently sitting on live.


jdalphon:
Awesome

Issue #2088, opened 2015-03-31
Group by contributors in vis


stowellm:
**General description:** There should be an option to group by contributors in a vis.  For example, if 5 users under contributor key "A", 7 users under contributor key "B", user "Little Timmy," and user "Little Billy" contribute data sets, the vis should have a group-by option for contributors that would list "Key: A, Key: B, User: Little Timmy, User: Little Billy."

**live/dev/localhost:** dev
**iSENSE Version:** v6.5
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** Ubuntu 14.04 LTS
**Browser/Version:** Chrome 41

**Steps to Reproduce:**
1. Make a project with some number fields.  Add location for pizzazz.
2. Add two contributor keys.
3. Contribute multiple data sets with the two keys and two additional users.
4. Visualize the data.  Start on Scatter, but it should work for all other relevant vises.
5. Note there should be a "Contributors" group-by option once this feature is complete.

**Associated Image(s):**
Once implemented, should look like this:
![lolthing](https://cloud.githubusercontent.com/assets/1039162/6941530/b72a9454-d84e-11e4-89de-a477a593cffa.png)

bdonald25:
Testers: Change the group by option to group by contributor. Make sure it works for new and old vises, new and old default vises, and new and old saved vises.

Also make sure that all the old saved vises and default vises are the same on live and dev. They should have all the same options for all their tools other than the option to group by contributor. If different things are checked off for a saved vis on live and dev then that is a bug.

jaypoulz:
@bdonald25 was this patch based on Mike's previous attempt? Because this is awesome! :+1: I took a quick look at the migration - gawd that is a monster. Sorry about that. I meant to fix this with 2.0, but our history of quick and dirty vis configurations made this quite a beast to tackle. Nice work!

bdonald25:
@jaypoulz It was loosely based on Mikes. I redid it from scratch without looking at his because Mike wasn't sure where the problem was but I noticed as I was almost done that mine was very similar to his, so I used his as reference to finish it off. His attempt was 90% of the way there, he had just skipped over some things that needed to be migrated.

jaypoulz:
I cannot :+1: this enough! :smile_cat:

pat-son:
PASSED testing on Firefox 44.0.2 (Mac OSX Yosemite 10.10.5).

bdonald25:
[...]

PR #2092, opened 2015-04-01
Added contributor group to vis

stowellm:
Addresses #2088 

I moved obtaining relevant vises to a new method because the `displayVis` method was too long.  I'm also bothered by the fact that there are functions both in camel and snake case in this file, so @jaypoulz, let me know which we should follow.

contrib


jdalphon:
This makes me so happy. +1 for @stowellm

app/controllers/visualizations_controller.rb
```diff
@@ -256,13 +265,19 @@ def displayVis
         arr.push "#{dataset.title}(#{dataset.id})"
         arr.push 'All'
 
+        # push unique contributors
+        if dataset.key.nil?
+          arr.push "User: #{User.find(dataset.user_id).name}"
```
jdalphon: For things like this we should be doing something more like
```ruby
User.select(:name).find(1).name
```
This will only instantiate the object 
```ruby
#<User id: nil, name: "James Dalphond"> 
```
instead of
```ruby
#<User id: 1, firstname: nil, lastname: nil, email: "USERS EMAIL", group_id: nil, validated: false, password_digest: "SOME LONG STRING", created_at: "2015-03-25 18:35:34", updated_at: "2015-03-25 18:35:34", validation_key: "ANOTHER LONG STRING", admin: false, hidden: false, bio: "", news_id: nil, last_login: "2013-08-16 12:00:00", name: "USERS NAME"> 
```
We have been running into issues of memory usage, and if we all start doing this wherever necessary, it will relieve some of the burden.

stowellm:
Thanks for the update @jdalphon - ready for re-review.

stowellm:
Scratch that - Jeremy and I noted that I will need to write a migration.  By adding a new virtual field, *every* field selection for vises (e.g. defaulting to some number for X-axis in a default vis) is offset by 1.

stowellm:
@jaypoulz / @jdalphon - I wrote the migration to perform what we need.  Would either (preferably both) of you mind reviewing it, ensuring I did not miss anything or am not doing anything wonk?  Warning ahead of time: there's a lot here, so the review will take time.

I will still need to test this to ensure it works as well.  How would you recommend I do so so that it doesn't break live when it gets pushed up?

Also worth noting: this migration is built off of Jake's regression patch (#2096) and thus **cannot be merged** until his PR is merged as well.

stowellm:
Well, I got a lot of Rubocop to fix - but @jaypoulz, remind me to talk to you sometime about JSON parsing and this migration.

stowellm:
Ping @jaypoulz - ready for you to merge locally and test when you get a chance :smile:

jaypoulz:
Saved vises be crashing *eeek* :cry: Especially bar and histogram. Idk about Scatter and Timeline because Jake broke those.

app/controllers/visualizations_controller.rb
```diff
@@ -209,7 +209,7 @@ def displayVis
     field_count = []
 
     # build list of datasets
-    if  !params[:datasets].nil?
+    if !params[:datasets].nil?
```
jaypoulz: how about unless params[:datasets].nil?

db/migrate/20150403162309_refactor_invisible_fields.rb
```diff
@@ -0,0 +1,308 @@
...
+        data = refactor_data(data, dir)
+        v.data = JSON.dump(data)
+      end
+
+      # save the new globals and data
+      v.save
+    end
+  end
+
+  def refactor_timeline(timeline, dir)
```
jaypoulz: timeline and scatter should call the same function

app/controllers/visualizations_controller.rb
stowellm: But there's an else statement :)

stowellm:
@jaypoulz - This just needs re-testing :smile:
- create project on your everyday branch
- switch to the branch that has this code on it
- run the migration (or if already ran, `rake db:migrate:redo VERSION=20150403162309`)
- confirm vises not be all wonk

jaypoulz:
Okay, so a couple things to note here:

1. I don't think you should commit the schema file in this case. This is because when we pull, rails assumes our db is up to date instead of running the migrate.
2. Bar breaks on sort-by (savedVis)
3. Summary vis field selection is broken (savedVis)
4. *Important* Project vises with only certain vises default seems to break the color mapping for all the other vises. Example: I only defaulted a scatter/timeline, and all the other vises are broken

stowellm:
- [ ] Summary magically introduces new data points (saved vis)

stowellm:
So.  Some things are fixed, but the checkboxes above that aren't checked are not.

jaypoulz:
Honestly, the right thing to do here is to re-factor the code so that it no longer relies on induces, but rather uses the id's we have already assigned fields. This would make it infinitely easier to track virtual fields, and would allow this (and similar features) to pass without a migration.

jaypoulz:
@bdonald25 this will never pass and tbh it would take a lot of work (and probably meetings with me) to explain what needs to be done.

bdonald25:
#2337 Replaces this pull request

Issue #2084, opened 2015-03-27
Tutorial console error


bdonald25:
**General description:** On any tutorial page. I get an error in the console "Uncaught TypeError: undefined is not a function"

**live/dev/localhost:** live
**iSENSE Version:** v6.5
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 41.0.2272.104

**Steps to Reproduce:** Any of the tutorial pages.

10

kcarcia:
TESTERS: Verify you get no console error on the tutorials page. Also verify you can post a tutorial.

stowellm:
PASSED testing on Chrome 41.0.2272.118 (Ubuntu 14.04 LTS).

JKinsman:
PASSED testing on Safari 8.0.3 (mac).

jdaigneau:
PASSED testing on Chrome 41.0.2272.118 (mac).

Issue #2083, opened 2015-03-27
News console error


bdonald25:
**General description:** While on the news page. Console gives an error telling me my chrome extension is invalid.

**live/dev/localhost:** live
**iSENSE Version:** v6.5
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 41.0.2272.104

**Steps to Reproduce:** News Page.

10

jdalphon:
@bdonald25 I get no such error.  Does anybody else see this error?

kcarcia:
TESTERS: Verify you get no console error on the news page. Also verify you can post a news article.

stowellm:
PASSED testing on Chrome 41.0.2272.118 (Ubuntu 14.04 LTS).

JKinsman:
PASSED testing on Safari 8.0.3 (mac).

jdaigneau:
PASSED testing on Chrome 41.0.2272.118 (mac).

PR #2103, opened 2015-04-09
Getting rid of useless ajax that only admins see. #2083 #2084

jdalphon:
We currently use crappy ajax to publish and unpublish news/tutorials causing: 
#2083 
#2084

Now it is simply a form with a hidden field set to be the opposite of the current hidden state.

Also tutorials were broken in that admins couldn't even see hidden ones. So that is fixed and the tests are better for it.


kcarcia:
@jdalphon code looks fine to me. Will test locally tomorrow morning.

kcarcia:
Those pages could probably use a little UI love, but since they are really admin only, that's low priority. Looks good to me.

jaypoulz:
@jdalphon I think maybe some of this ajax was preventing XSS attacks on news and tutorials. Salani was breaking dev by injecting scripts into news article titles then publishing them to the homepage. Do you mind adding the rails sanitization call to that presave? Should be a few lines, but you know it best

jdalphon:
@jaypoulz There was no ajax on either of those pages that did anything but handlebars for changing the hidden property of the element. Please open a new issue to fix this bug. 

It would be pretty low priority as we would not be injecting scripts into our own website. You need to be admin to create/edit tutorials or news.

Issue #1794, opened 2014-09-26
Deleting a user doesn't delete projects or saved vises


asalani93:
**General description:** Deleting a user doesn't delete projects or saved vises.  You can't see the projects or vises from the index pages, but the URL still works and they're still in the user's page because #1773.  Also, it seems as if media objects are *sometimes* deleted.  For example, this project:
http://rsense-dev.cs.uml.edu/projects/838
had it's media object deleted, but there was another one that still has it's media object.

**live/dev/localhost:** dev
**iSENSE Version:** 5.7.74
**Logged in (Y or N):** Y/N
**Admin (Y or N):** Y/N

**OS:** Ubuntu 14.04.1 x64
**Browser/Version:** Chrome 36

**Steps to Reproduce:**
1. Create a user.
2. Create a project
3. Give that project a dataset or two.
3. Create a saved vis from that project.
4. Give each a media object.
5. Delete the user.
6. Observe as the projects and vises are still there, and the media objects may or may not still be there.

**Associated Image(s):**
nah

stowellm:
Kind of goes with #1773

kcarcia:
TESTERS: 
If you delete a user, please verify
(1) their empty projects should just be deleted
(2) their projects with data sets only contributed by him/her should be deleted
(3) their projects with data sets contributed by other users will not be deleted; the owner will be admin

This is how this works:
"You still can't delete a project with data sets by default. If you delete a user, all of his/her data sets are deleted first. If it is found that now his/her project can be deleted (i.e. only that user contributed to it), it gets deleted. Otherwise, since that project has contributions from other users, ownership switches to admin. Then the user gets deleted. We should note only admins can delete users though."

stowellm:
PASSED testing on Chrome 42.0.2311.135 (linux).
DISREGARD THIS IS JUST A TEST

JKinsman:
PASSED testing on Chrome 42.0.2311.135 (Mac OSX 10.10.2).

jdaigneau:
PASSED testing on Chrome 42.0.2311.90 (mac).

PR #2112, opened 2015-04-13
When you hit delete, it actually deletes

stowellm:
Addresses #1794 and any other issue we have regarding deleting things.

@jdalphon, @jaypoulz, and @kcarcia - No idea who may be interested in this pull request, but review/test at your leisure.  There is no more "hiding" things when we delete them.  They actually get deleted.


kcarcia:
@stowellm How does this work? Does a deleted user's everything really go away?

stowellm:
@kcarcia and I talked about this in person, but I'll summarize the discussion:

Deleting a user deletes everything in order of importance, e.g. news/likes/tutorials get deleted first.  Data sets are wiped second to last, and finally projects.  When attempting to delete a project, the code will check if any data sets still exist (recall this user's data sets have already been deleted, so a project created by this user with only his/her data sets previously now has none).  If data sets exist, the project ownership is changed to admin. Otherwise it is deleted.

stowellm:
@jdalphon Do you mind reviewing this to make sure it's sensible for what you believe should happen?

stowellm:
Bump.

stowellm:
speed_bump_graphic

jdalphon:
@stowellm Sorry, I was not getting emails about this issue for some reason. 

Yes, this looks sane for now. There are some other implications that we may need to deal with later. We may want to look into how other services like google docs, dropbox, and others deal with deleting shared content that has edits by other people. 

I'm still not convinced that deleting stuff out from under users without at least notifying them is a good idea. 

What is the current behavior for deleting a project outright?

stowellm:
Currently on dev:
You can't delete a project with data sets.  Deleting a user that owns a project with data sets, however, just hides the project.  You can only visit it by direct URL.  The user will still be the "deleted" user since only the user's email gets changed.

In this pull request:
You still can't delete a project with data sets by default.  If you delete a user, all of his/her data sets are deleted first.  If it is found that now his/her project can be deleted (i.e. only that user contributed to it), it gets deleted.  Otherwise, since that project has contributions from other users, ownership switches to admin.  Then the user gets deleted.  We should note only admins can delete users though.

Another note:
As it stands, a project owner can already delete everyone else's data sets without warning, then delete his/her own project.

jdalphon:
@stowellm That last part is what I was worried about. But if the owner can delete peoples stuff at any time, then is this just a moot point? 

I would say your PR as right now is fine, but we may change our minds in the next big meeting and say project owners are god and can delete their stuff. We will see after the meeting

stowellm:
Sounds good - @kcarcia, ready for you to pull.

kcarcia:
@stowellm I will pull this tomorrow after testing is complete. :)

Issue #2114, opened 2015-04-15
Edit fields page: can add multiple timestamps


stowellm:
**General description:** Made a project, went to add fields, incremented the amount of numbers I wanted to add to "10."  Then I decided, "No, I'm actually going to hit 'Add Timestamp' now."  It added 10 timestamps.

**live/dev/localhost:** dev (any)
**iSENSE Version:** 6.5
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu LTS 14.04
**Browser/Version:** Chrome 41

**Steps to Reproduce:** See description.

**Associated Image(s):**
![2manytiemstmpas](https://cloud.githubusercontent.com/assets/1039162/7159255/e33714f4-e34e-11e4-8221-abe5c9d77b33.png)

kcarcia:
TESTERS: You shouldn’t be able to add multiple timestamps (or location fields) to the table.

bdonald25:
PASSED testing on Chrome 42.0.2311.135 (mac).

JKinsman:
PASSED testing on Chrome 42.0.2311.135 (Mac OSX 10.10.2).

jdaigneau:
PASSED testing on Chrome 42.0.2311.90 (mac).

jdaigneau:
PASSED testing on Chrome 42.0.2311.90 (mac).

PR #2120, opened 2015-04-27
New Edit Fieilds Table

kcarcia:
#2009 
#2114 

- Table dynamically created with coffeescript.
- No ajax.
- Form only submits on save.
- Cancel button to go back.

:shipit:


stowellm:
:+1: for correctly spelling fieilds

JKinsman:
@kcarcia I pulled the branch and tested locally.  Great work! :smile: 

I did, however, notice a couple minor issues:
1.  If the new fields fail to save (ex: duplicate names) it clobbers all the fields I just tried to add.  Would it be possible to save the state, then repopulate the table on failure?
2.  The error message on fail is not very helpful.
3.  Text fields, like number fields, should be given unique names by default.  Adding two numbers gives the fields 'Number', and 'Number_2'.  Since the user can have an unlimited number of text fields, those should probably do the same.
4.  Personally, I think the table is a little too flat visually.  Maybe add a gentle background color to a row based upon the type of field, or something?

I did not review the code, although I'd be happy to if you'd like. :blush:

stowellm:
Comments:
- See @JKinsman comment 3.  Also doesn't make unique number names.
- I made my field names unique, hit save, and got an error "Field names must be unique".  They were unique.  I hit save again and it worked.

kcarcia:
Jake's Feedback:
- The controller immediately returns if there's an error, so the table that's loaded contains only the fields that have been successfully saved. Populating the table with the error rows would likely require some more backend work that extends past my time here. Great suggestion, though.
- Was there a particular error message that was confusing?
- Going to try and rewrite a ruby method in Coffeescript to make this happen.
- The striped table is confusing because the table changes colors as rows are added. I added the regular table style back and made the header grey though.

Mike's Feedback:
- See above. ^
- We actually determined we can't reproduce this.

Thank you both!!!

kcarcia:
So I re-implemented the whole Number_# Text_# thing. Wooohoo! So this is ready for review now @jaypoulz. I would've liked to spend some more time simplifying my code, but I ran out of time. Definitely let me know if you have suggestions.

app/views/projects/edit_fields.html.erb
```diff
@@ -1,77 +1,79 @@
...
+    <div class="col-lg-8 col-lg-offset-2">
+      <h1>Setup Fields Manually</h1>
+      <div class="center">
+        <div class="col-sm-3 col-xs-6 add_fields">
+          <button value="Number" class="btn btn-success center new_field" id="number">Add Number</button>
+          <p class="description">Any number value. "Units" specifies how the value is expressed. e.g. 'Temperature' could have a unit of 'C'</p>
+        </div>
+        <div class="col-sm-3 col-xs-6 add_fields">
+          <button value="Text" class="btn btn-success center new_field" id="text">Add Text</button>
+		  <p class="description">Any string value. "Restrictions" are a comma separated list of values that limit input to only predefined values.</p>
```
jaypoulz: EEEKKKKKK tabbbbbbb :fearful:

app/assets/javascripts/projects/edit_fields.js.coffee
```diff
@@ -1,78 +1,152 @@
...
+      addRow(["""<input class="input-small form-control" type="text" name="latitude"
+                 value="Latitude">""", "Latitude", "deg", "", """<a href="#" fid="0"
+                 class="field_delete"><i class="fa fa-close slick-delete"></i></a>"""])
+      document.getElementById('location').disabled = true
+
+    # Delete field, enable timestamp/location buttons (NOTE: fid is 0 when the field
+    # hasn't yet been added to project in database)
+    $('#fields_table').on 'click', '.field_delete', ->
+      # fid of row being deleted
+      fid = $(this).closest('a').attr('fid')
```
jaypoulz: anytime you have `this` in coffeescript, replace with `@`

app/assets/javascripts/projects/edit_fields.js.coffee
```diff
@@ -1,78 +1,152 @@
...
+
+      # row index of row being deleted
+      row_index = $(this).closest('tr').index()
+
+      # row name of row being deleted
+      row_name = $(this).closest('tr').attr('name')
+
+      # Decrease counter based on row name
+      if row_name == 'latitude' || row_name == 'longitude'
```
jaypoulz: technically this should be a switch case, but we can let it slide

app/assets/javascripts/projects/edit_fields.js.coffee
```diff
@@ -1,78 +1,152 @@
...
+                 class="field_delete"><i class="fa fa-close slick-delete"></i></a>"""])
+      document.getElementById('location').disabled = true
+
+    # Delete field, enable timestamp/location buttons (NOTE: fid is 0 when the field
+    # hasn't yet been added to project in database)
+    $('#fields_table').on 'click', '.field_delete', ->
+      # fid of row being deleted
+      fid = $(this).closest('a').attr('fid')
+
+      # row index of row being deleted
```
jaypoulz: Can the r in row be capitalized

app/assets/javascripts/projects/edit_fields.js.coffee
```diff
@@ -1,78 +1,152 @@
...
+      row_index = $(this).closest('tr').index()
+
+      # row name of row being deleted
```
jaypoulz: Again here

jaypoulz:
Just these small fixes, and all snake case var names should be camelCase in CoffeeScript -- that concludes the style review

jaypoulz:
:+1: for :shipit: 

Just one thing if it's an easy fix:
- [ ] If you have the same field name, the error message comes back like `["Name has already been taken"]`, but I don't think the brackets should be there

Issue #2024, opened 2015-02-27
Field restrictions should default in database as empty array


jaypoulz:
**General description:** Field restrictions should default to empty array

**live/dev/localhost:** live
**iSENSE Version:** v6.15-24-g6a35ce4
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** X11
**Browser/Version:** Firefox 37.0

**Steps to Reproduce:** Migration 20141015152508_remove_whitespace_from_restrictions brought attention to the fact that there were actually 3 kinds of representations for empty field restrictions.  It would be nice if we changed this to a singular representation. The representation chosen by the new version of the migration is empty array. See #2022 for more details.

9

bdonald25:
testers: Add restrictions for a text field on a project. Make sure the restrictions are working correctly.

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

PR #2130, opened 2015-05-13
Fixes field restrictions

asalani93:
Addresses #2024.

Field restrictions are now constrained to being just an array of restrictions.  Previously, a lack of restrictions could be represented as an empty string, an empty array, or nil.

This was done by having the fields model actually validate restrictions.  I also had to fix some tests, and I removed some code from FileUploader that would reject any text entered in a field that had no restrictions because the entered text wasn't in the empty array of restrictions.

I also cleaned up how validation was done.  I am also no longer returning just one error if the user fails create a field, but instead all of them.  This useful in regards to the API and stuff.

### DIFF
--- a/app/models/field.rb
+++ b/app/models/field.rb
@@ -2,7 +2,7 @@ include ApplicationHelper
 class Field < ActiveRecord::Base
   validates_presence_of :project_id, :field_type, :name
   validates_uniqueness_of :name, scope: :project_id, case_sensitive: false
-  validates_with RestrictionsValidator, fields: [:restrictions]
+  validate :validate_restrictions
   belongs_to :project
   serialize :restrictions, JSON
   alias_attribute :owner, :project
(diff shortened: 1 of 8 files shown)


app/models/field.rb
```diff
@@ -50,37 +56,38 @@ def self.get_next_name(project, field_type)
...
+  def validate_values
+    # verify that restrictions is an array of strings
+    if !restrictions.is_a? Array
+      errors.add :restrictions, 'must be in an array'
+    elsif !restrictions.reduce(true) { |a, e| a and e.is_a? String }
+      errors.add :restrictions, 'must all be strings'
+    end
+
+    if !field_type.nil? and ![1, 2, 3, 4, 5].include? field_type
+      errors.add :field_type, 'must be a number between 1 and 5'
```
bdonald25: If you entered numbers for restrictions it would fail.

bdonald25:
Looks good to me other than that one thing.

Issue #2127, opened 2015-05-09
histogram tooltip should show exact value of sample (not bin-rounded value)


fgmart:
The tooltip popup on histogram does a lovely job highlighting that you're inspecting exactly one sample.

But, I'd like it to show the exact value of the sample that has been binned. Right now, it seems to round off to fit into the bin.

See below, from http://isenseproject.org/projects/395/data_sets/8618.

![screen shot 2015-05-09 at 11 46 51 am](https://cloud.githubusercontent.com/assets/1038856/7550888/5253e7fe-f641-11e4-9b29-79f7b78fc2ff.png)

asalani93:
TESTERS: However over values and make sure that it shows both the bin the block corresponds to, and its value.

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

PR #2131, opened 2015-05-13
Histogram now shows real numbers instead of binned numbers

asalani93:
Addresses #2127.

Very easy fix.  Now histogram shows the actual recorded value over the value that results from binning it.  Only does things on vises that actually show individual records.


jaypoulz:
This is good.

Issue #2128, opened 2015-05-09
browser window title capitalizes name of data set when it should not


fgmart:
See attached. In the tab, the name of the window/dataset is "Ps...".  It should be "ps...".

![screen shot 2015-05-09 at 11 57 00 am](https://cloud.githubusercontent.com/assets/1038856/7550906/85fa1fe6-f642-11e4-8e39-1f2d8e72d896.png)

bdonald25:
Testers: Go to a vis for a single data set. Make sure the data set name in the browsers tab is not capitalized.

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

PR #2135, opened 2015-05-14
no longer capitalize data set title in tab

bdonald25:
#2128


app/helpers/application_helper.rb
```diff
@@ -133,7 +133,7 @@ def render_title
         # viewing data sets
         if @datasets.count == 1
           # viewing a single data set
-          @datasets[0].title.capitalize.html_safe
+          @datasets[0].title.html_safe
```
asalani93: remove html_safe

Issue #2110, opened 2015-04-10
Picture uploaded multiple times in project description


jdaigneau:
General description: When adding a picture to the description of a project, I hit save and the page didn't update, so I clicked it once more. The page then loaded and in the media section on the right side of the page, the picture had uploaded three times.

live/dev/localhost: dev
iSENSE Version: v6.6
Logged in (Y or N): Y
Admin (Y or N): N

OS: Mac OSX 10.9.4
Browser/Version: Chrome 41.0.2

Steps to Reproduce: Add picture to project description. Hit save multiple times.

kcarcia:
@JKinsman @jaypoulz @jdalphon So this is real and live. If you click save multiple times when you're editing a description w/ a media object, the media object is saved multiple times. No idea what's causing this. Anyone have any ideas off the top of their head? (Mostly looking at you Jake since you know Summernote)

asalani93:
TESTERS: Try to get two of the exact same images into a project, visualization, or anything that can have images uploaded to it.  Also, try other media objects.

KatherineBrunelle:
FAILED testing on Safari 8.0 (mac).
Pictures don't upload at all.

asalani93:
FAILED testing on Chrome 42.0.2311.90 (linux).
Pictures don't upload, old duplicates still exist

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2137, opened 2015-05-15
Prevent duplicate images from being uploaded to the same project/vis/etc...

asalani93:
Addresses #2110, #2139.

Prior to this, users could upload duplicate media objects.  This makes no sense if the files are exactly the same.  It was also very easy to upload duplicate media objects if you added one to the project description and then clicked save twice by accident.

This pull request:
* gives each media object a hash
* removes all non-featured duplicate media objects
* prevents duplicate media objects from being uploaded
* prevents duplicate media objects from being created through summernote
* for duplicate images in summernote, the correct URL from the already-uploaded image is found and then put in place of the duplicate media object.

ALSO:
* you can now no longer upload media objects to any project or data set you feel like - instead, you need to actually be allowed to edit it


asalani93:
Now addresses entirely unrelated issue #2139.

app/models/media_object.rb
```diff
@@ -35,6 +36,40 @@ def sanitize_media
...
+    elsif !data_set_id.nil?
+      owner = DataSet.find data_set_id
+    elsif !tutorial_id.nil?
+      owner = Tutorial.find tutorial_id
+    elsif !visualization_id.nil?
+      owner = Visualization.find visualization_id
+    elsif !news_id.nil?
+      owner = News.find news_id
+    else
+      nil
```
bdonald25: Return here and remove
 if owner.nil?
     return
  end

bdonald25:
Good to Pull

Issue #2005, opened 2015-02-13
Field matching suggests a text field for timestamps


JKinsman:
**General description:** If you upload fields for a project from a file, timestamp fields will be matched as text fields.

**live/dev/localhost:** dev
**iSENSE Version:** 6.1
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Mac OSX 10.10.2
**Browser/Version:** Safari 8.0.3

**Steps to Reproduce:**
1.  Use a CSV with a timestamp field to create fields for a project (template.csv is fine)
2.  Field matching will suggest it is a text field.

**Associated Image(s):**

kcarcia:
@jdalphon Ok so as it turns out, when you export csv and xls files from Google drive, it changes the timestamp format to mm/dd/yyyy, and this explains why timestamp matching is not working. Does this mean we should extend the regex to recognize dates in that format?

kcarcia:
So this is a Google problem.

However, it would be nice if we could allow users to specify the format of their timestamps on the field matching form. So say their data is in the form mm/dd/yyyy... we would take that information to field match and convert their timestamp to the format we require for parsing. That way we avoid parsing hell, and everyone wins.

This would be a nice-to-have feature but is low priority.

bdonald25:
Testers: Upload data with time formats mm/dd/yyyy and dd/mm/yyyy and make sure they both work as expected.

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

PR #2138, opened 2015-05-18
Quick fix for timestamp field matching

asalani93:
Address #2005.

This is just a quick band-aid for supporting multiple timestamp formats.

### DIFF
--- a/lib/file_uploader.rb
+++ b/lib/file_uploader.rb
@@ -212,13 +212,14 @@ class FileUploader
     types['latitude'] = []
     types['longitude'] = []
 
-    regex = %r{((?<year>\d{4})(-|\/)(?<month>\d{1,2})(-|\/)(?<day>\d{1,2})|([uU]\s\d+))}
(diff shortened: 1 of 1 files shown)


jdalphon:
@asalani93 As long as this actually is just a band-aid and does not stay around long.

asalani93:
It'll get fixed in the eventual future, but not right now.  Too many features that are actually important and have higher beneficiality:code ratios.  Will be done soon though.

jdalphon:
@asalani93 Did you test to make sure both get uploaded correctly?

asalani93:
yes

jdalphon:
This will accept
01/02/2015 to mean Febuary 1st or January 2nd which is why we had one
STRICT date format. Which is why we should
have a format picker sooner rather then later on file upload.

On Mon, May 18, 2015 at 2:16 PM, Anthony Salani <notifications@github.com>
wrote:

> yes
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/isenseDev/rSENSE/pull/2138#issuecomment-103154899>.
>



-- 
--------------------------------
James Dalphond
UML Computer Science
Engaging Computing Group

Issue #1896, opened 2014-12-19
Error bars on bar chart


fgmart:
**General description:**
It is commonly used in statistics to display error bars on summary bar charts.
E.g, you have a bar that collapses many values into a mean. You get a error bar with one std dev above and one below.

Or the error bar shows the 75th percentile (above mean) and 25th percentile below mean.

This would be a nice feature.

**live/dev/localhost:** N/a
**iSENSE Version:** N/a
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** N/a
**Browser/Version:** N/a

**Steps to Reproduce:** N/a

**Associated Image(s):** None

stowellm:
@jaypoulz and/or @fgmart  - To work on this I'll need a face-to-face conversation explaining what we want here and whether or not it is within the scope of my current abilities to work on this.

jaypoulz:
@stowellm Best way to learn vis is just to start doing it :smile_cat: but yes, we will talk in person

asalani93:
TESTERS: Visualize some bar charts and observe that "Mean with Error" shows bar charts with a standard deviation whisker plot thing.  Also, verify that these error bars do not get drawn on pie chart.

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
FAILED testing on Firefox 40.0 (linux).
Breaks for negative (or 0) on logarithmic access

bdonald25:
New bug addressing the problem Jeremy discovered. #2158

bdonald25:
Testers: Visualize some bar charts and observe that "Mean with Error" shows bar charts with a standard deviation whisker plot thing. Also, verify that these error bars do not get drawn on pie chart.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2145, opened 2015-05-21
Adds standard deviation error bars to bar chart

asalani93:
Addresses #1896.  We now have error bars on bar chart.

This was added as an additional analysis type ("Mean with Error").  However, it makes no sense to show error bars on pie chart.  Therefore, I had to add a system so that different visualizations that use analysis types can have what analysis types they use be customized on a per-vis level.

### DIFF
--- a/app/assets/javascripts/visualizations/highvis/baseVis.coffee.erb
+++ b/app/assets/javascripts/visualizations/highvis/baseVis.coffee.erb
@@ -420,7 +420,7 @@ $ ->
       Draw tool controls (Analysis Type, Log Y Axis, etc.) for Pie Chart and
       Bar Charts
       ###
-      drawToolControls: (sortBy = false, logYAxis = false, analysisTypes = []) ->
+      drawToolControls: (sortBy = false, logYAxis = false, analysisTypeExcludes = []) ->
         inctx = {}
 
         # Populate sort fields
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/visualizations/highvis/baseVis.coffee.erb
```diff
@@ -36,17 +36,18 @@ $ ->
...
-      ANALYSISTYPE_MAX:       1
-      ANALYSISTYPE_MIN:       2
-      ANALYSISTYPE_MEAN:      3
-      ANALYSISTYPE_MEDIAN:    4
-      ANALYSISTYPE_COUNT:     5
+      ANALYSISTYPE_TOTAL:      0
+      ANALYSISTYPE_MAX:        1
+      ANALYSISTYPE_MIN:        2
+      ANALYSISTYPE_MEAN:       3
+      ANALYSISTYPE_MEAN_ERROR: 4
```
jaypoulz: This is not compatible with saved/default regressions. Anything saved with analysis type 4 will now have mean_error instead of median.

jaypoulz:
As mentioned in review in person - the array of data.mean passed to getGroupedData is unused, and the arrays of analysis types should be defaulted to empty and use negative logic to skip types instead of listing all those you should loop on.

jaypoulz:
This is way prettier. Thanks for making that fix!

Issue #2146, opened 2015-05-21
Data set edit page misbehaving


asalani93:
**General description:** The data set edit page has the following problems:
* Loading a data set  with more than ten rows appends an additional #rows - 9 to the bottom of the data set.
* The active cell is by default somewhere in the middle of the page.  Not sure why.  In conjunction with the previous issue, this leads to the data set entry page to look seemingly blank.
* Null values are left in the data set and not converted to empty strings.  This prevents the data set from being saved.
* Uploading a file with missing data leaves one or more nils in the data set.  In conjunction with the issue above, this prevents data sets that have been uploaded from being saved again.  It also gives uploaded data sets the appearance that they have no data in them when you go to edit them.

The last one technically isn't an issue with the data set edit page, but with the FileUploader class.

This issue report is solely for documenting the bug for the impending pull request.

This is the problem Jim came across when setting up the data for the boat house.

**live/dev/localhost:** yes
**iSENSE Version:** v7
**Logged in (Y or N):** Y
**Admin (Y or N):** Y/N

**OS:** Ubuntu 14.04 x64
**Browser/Version:** Chrome 42 x64

**Steps to Reproduce:**
* Upload data from a CSV.  To view the entirety of the bug, upload a large CSV (around 40 rows or more) with some values left blank.
* Go to edit the data set.  Observe the plethora of empty rows, and your scroll location.  Note how easily this could be confused with not having any data uploaded.
* Try and save the data set.  Observe how you can't.  Note how this is generally not desirable behavior.

**Associated Image(s):**
![missing-datapng](https://cloud.githubusercontent.com/assets/7584534/7753018/c9393d7e-ffb0-11e4-974d-03696e6e6fe4.png)

bdonald25:
Testers: Edit a data set -> make sure everything works as expected.

KatherineBrunelle:
FAILED testing on Safari 8.0 (mac).
Clears text fields when highlighted. Also, won't let you save if you add data points, the save button says "saving" and freezes there.

asalani93:
FAILED testing on Chrome 42.0.2311.90 (linux).
Text editing fields are emptied upon selecting them.

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

bdonald25:
New issue addressing slick grid's remaining problems. #2157

PR #2147, opened 2015-05-21
Making the data set edit page behave

asalani93:
Addresses #2146.

There's a handful of issues that affected data set editing.  All are documented in the issue and all have been fixed.  I'm not sure if anything changed, or if the science of rivers data set was just the perfect storm for breaking data set editing.

### DIFF
--- a/app/assets/javascripts/data_sets/editTable.js.coffee
+++ b/app/assets/javascripts/data_sets/editTable.js.coffee
@@ -39,9 +39,6 @@ class Grid
   actions: []
 
   constructor: (@cols, @data, @submit) ->
-    console.log JSON.stringify @data
-    console.log @data
-
     # this is needed because slickgrid opens after this function completes
     @initialize()
     @subscribeEvents()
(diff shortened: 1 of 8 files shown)


bdonald25:
@asalani93 
Tab cycles through the cells and wraps around at the end of the row. Shift + Tab cycles back through the cells in the row but does not wrap back around.
In my opinion, the arrow keys should not cycle through the cells, they should move the cursor within the cell that you are editing. Especially since you have tab and shift+tab to cycle through the cells.

app/assets/javascripts/data_sets/editTable.js.coffee
```diff
@@ -17,6 +17,74 @@ uploadSettings =
   successEntry: (data, textStatus, jqXHR) ->
     window.location = data['displayURL']
 
+
+#      if e.keyCode == 13 and @grid.getDataLength() - 1 == @grid.getActiveCell().row
```
bdonald25: Remove commented out code.

Issue #2064, opened 2015-03-13
Flashing Red Error Indication Removal


kcarcia:
**General description:** Remove all remnants of flashing red errors. Like on the edit a data set title.

**live/dev/localhost:** live
**iSENSE Version:** v6.3
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Macintosh
**Browser/Version:** Chrome 41.0.2272.89

**Steps to Reproduce:** Example:
1. Go to a project with a data set.
2. Edit the title of that data set.
3. Type in <strong></strong>
Notice the flashing red.

1

bdonald25:
Testers: Try and find flashing red errors by entering invalid values in various places on the site. (enter characters for numbers, leave things blank, ect.)

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

PR #2148, opened 2015-05-22
Removes flashing red error indicators

asalani93:
Addresses #2064.

Removes flashing red text fields and replaces them with popovers saying what went wrong.


bdonald25:
@asalani93 One second is too quick for the error messages.

bdonald25:
Good to merge. Three seconds is much better than one.

Issue #2143, opened 2015-05-21
Project and vis search boxes are cut off on mobile.


bdonald25:
**General description:** Project and Vis search boxes are cut off in mobile.

**live/dev/localhost:** live
**iSENSE Version:** v7.0
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** Macintosh
**Browser/Version:** Chrome 42.0.2311.152

**Steps to Reproduce:** Go to project of vis index pages.

![screenshot_2015-05-20-23-45-06](https://cloud.githubusercontent.com/assets/4121924/7741266/e0176658-ff4a-11e4-81f8-389db9975116.png)

393

bdonald25:
Testers: Go to http://rsense-dev.cs.uml.edu/projects and http://rsense-dev.cs.uml.edu/visualizations on your phone. Make sure they look reasonable.

KatherineBrunelle:
FAILED testing on Safari 8.0 (mac).
In data visualizations, the project name and user name trail off the page. You can still scroll to read it, but it's off the main page. The project search box is fully functional.
You cannot see the search box on the table visualization.

bdonald25:
Kathrine's bug is a separate issue from this. Opened another issue #2156

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

PR #2149, opened 2015-05-26
Makes mobile view for projects index page look semi-reasonable

asalani93:
Addresses #2143.

I made some things look different for the project search on different screen sizes.  Search bar now has a search icon instead of a button that says search.


bdonald25:
@asalani93 Vis index page has the same search box. You should redo that one as well.

bdonald25:
Looks much better :+1:

Issue #1753, opened 2014-08-26
Large spaces between buttons on projects#show when logged out


asalani93:
**General description:**
There's only two buttons when you view a project while logged out, and there's a huge space between the two.

**live/dev/localhost:** live/dev/localhost
**iSENSE Version:** 5.6
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** Ubuntu 14.04
**Browser/Version:** Chrome 36 something

**Steps to Reproduce:**
* log out
* view a project
* observe spaces between buttons

**Associated Image(s):**
![screen shot 2014-09-10 at 11 31 04 am](https://cloud.githubusercontent.com/assets/1798956/4220169/862f83d0-38ff-11e4-8855-5d8ebec7c705.png)

asalani93:
TESTERS: Look at the image above and ensure that it looks worse than what is currently there.

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

PR #2150, opened 2015-05-27
this was a very tough bug to fix

asalani93:
Addresses #1753.

I moved some things around and now it works.


jaypoulz:
My only nitpick is that there is no space between project.id and %>.

Issue #2060, opened 2015-03-13
iSENSE does not support negative numbers


asalani93:
**General description:** Negative numbers cause the heatmap color to turn red.  Also, pie charts don't show if you have negative numbers.

**live/dev/localhost:** live
**iSENSE Version:** v6.3
**Logged in (Y or N):** Y
**Admin (Y or N):** Y

**OS:** X11
**Browser/Version:** Chrome 40.0.2214.115

**Steps to Reproduce:** I had negative values in my heatmap, and certain areas turned bright red.  It looks weird.

![negative-heatmap](https://cloud.githubusercontent.com/assets/7584534/6642169/0d351534-c979-11e4-8aaf-09a3bca5eb12.png)

stowellm:
Can confirm.
![canconfirm](https://cloud.githubusercontent.com/assets/1039162/6642208/5201dcc4-c979-11e4-9a09-a32268f88ffe.png)

bdonald25:
Testers: Look at a heat map for a field that has negative numbers.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2152, opened 2015-05-28
Reworked negative number support in Map and Pie

jaypoulz:
#2060


bdonald25:
Good once passes Travis

Issue #915, opened 2013-10-27
Uploadable zip files


AlanRosenthal:
We should be able to upload a zip file of datasets (where each file is a dataset).

The reason for this is that you can download a zip file of datasets, so it only makes sense that you can upload it again.

bdonald25:
Testers: Upload zip files with valid data sets make sure everything works. Also, try to break it with invalid files such as images, ect.

asalani93:
PASSED testing on Chrome 42.0.2311.90 (linux).

jaypoulz:
PASSED testing on Firefox 40.0 (linux).

AlanRosenthal:
About time

Evana13G:
LMAO alan @AlanRosenthal

PR #2153, opened 2015-05-28
Support for uploading zip files.

bdonald25:
You can now upload zip files. 
#915


app/controllers/data_sets_controller.rb
```diff
@@ -243,50 +243,72 @@ def export
...
-          format.html { render action: 'dataFileUpload' }
+          flash[:error] = "Data could not be saved: #{sane[:msg]}"
+          format.html { redirect_to project }
         end
-      end
-    else
-      respond_to do |format|
-        flash[:error] = "Data could not be saved: #{sane[:msg]}"
-        format.html { redirect_to project }
+        return
```
asalani93: I don't think this `return` does anything.

app/controllers/data_sets_controller.rb
```diff
@@ -305,16 +327,59 @@ def dataFileUpload
       end
     end
 
-    if params[:file]
-      @filename = DataSet.get_next_name(project, params[:file].original_filename.split('.')[0])
-    else
-      @filename = DataSet.get_next_name(project, 'Google docs')
+    if params[:file] && params[:file].original_filename.split('.')[1] == 'zip'
```
asalani93: might have problems with filenames like `dogs.cats.zip` -> will extract `cats` as the filetype, not `zip`.  Try `-1` instead of `1`.

app/controllers/data_sets_controller.rb
```diff
@@ -305,16 +327,59 @@ def dataFileUpload
...
+
+              uploader = FileUploader.new
+              data_obj = uploader.generateObject(tempfile)
+              results.push(uploader.match_headers(project, data_obj))
+
+              filenames.push(DataSet.get_next_name(project, entry.name.split('/')[1].split('.')[0]))
+
+            rescue Exception
+              flash[:error] = "Error reading file #{entry.name}"
+              next
```
asalani93: Shouldn't this rescue handle the cause of `__MACOSX` and `DS_Store` files?  There seems to be no reason to special case those.

app/controllers/data_sets_controller.rb
```diff
@@ -305,16 +327,59 @@ def dataFileUpload
...
+            end
+            begin
+              tempfile = Tempfile.new(entry.name.split('/')[1])
+              tempfile.write(entry.get_input_stream.read)
+
+              uploader = FileUploader.new
+              data_obj = uploader.generateObject(tempfile)
+              results.push(uploader.match_headers(project, data_obj))
+
+              filenames.push(DataSet.get_next_name(project, entry.name.split('/')[1].split('.')[0]))
```
asalani93: Same as above re: `1` and `-1`.

asalani93:
Just some small things that might not even be problems.

Issue #2161, opened 2015-06-01
On chrome 43, vises are sometimes cut in half.


bdonald25:
**General description:** On chrome 43, vises sometimes are cut in half.

**live/dev/localhost:** live
**iSENSE Version:** v7.1
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** X11
**Browser/Version:** Chrome 43.0.2357.81

**Steps to Reproduce:** Looking at vises.

393

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

PR #2162, opened 2015-06-01
fix visualizations from cutting in half on chrome 43

bdonald25:
#2161


app/assets/stylesheets/visualizations.css.scss
```diff
@@ -301,7 +301,7 @@ body[data-page-name='visualizations/show'] .visualizations-controller,
 
   .vis-canvas {
     height: inherit;
-    width: inherit;
+    width: auto;
```
asalani93: The default value for width is auto, so you can just delete it.

Issue #2158, opened 2015-05-29
Error bars on bar chart don't when data is logarithmic and negative


asalani93:
**General description:** See title

**live/dev/localhost:** dev
**iSENSE Version:** 7.1
**Logged in (Y or N):** Y/N
**Admin (Y or N):** Y/N

**OS:** Ubuntu 14.04 x64
**Browser/Version:** Chrome 42 x64

**Steps to Reproduce:**
* Visualize a bar chart with error bars such that the analysis type would end up displaying negative numbers
* Observe as High Charts throws a fit trying to plot it

**Associated Image(s):**

bdonald25:
1. Visualize "Mean with Error" on a bar chart with negative numbers. Verify that it works the same as with positive numbers. 

2. Visualize "Mean with Error" on a bar chart that contains a logarithmic y-axis (you set this under tools). The error bar should only be shown on the top half of the graph.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2165, opened 2015-06-01
Standard deviation on bar chart again

asalani93:
Addresses #1896, #2158.

Fixes the migration to test a little more rigorously for what data is actually present in saved and default vises.  Turns out it was less consistent than I thought it was.

Also fixes standard deviation on logarithmic y-axis.  According to [this article](http://www.graphpad.com/faq/file/1487logaxes.pdf), showing standard deviation error bars on a log y axis is considered an "abuse" and makes literally no sense, so I made my fix reflect that.  Only the top part of error bars for non-negative means are shown.

### DIFF
--- a/app/assets/javascripts/visualizations/highvis/bar.coffee
+++ b/app/assets/javascripts/visualizations/highvis/bar.coffee
@@ -117,8 +117,13 @@ $ ->
               else
                 mean = groupedData[fid][gid]
                 innerStd = barData.map((x) -> (x - mean) ** 2).reduce((x, y) -> x + y)
-                stdDev = Math.sqrt (1 / (barData.length - 1)) * innerStd
-                [mean - stdDev, mean + stdDev]
+                stdDev = Math.sqrt((1 / (barData.length - 1)) * innerStd)
+                if globals.configs.logY != true
+                  [mean - stdDev, mean + stdDev]
+                else if mean > 0
+                  [mean, mean + stdDev]
+                else
+                  []
 
             @chart.addSeries errors
 
@@ -129,6 +134,7 @@ $ ->
 
       drawControls: ->
         super()
+
         @drawGroupControls(data.textFields)
         @drawYAxisControls(globals.configs.fieldSelection,
           data.normalFields.slice(1), false)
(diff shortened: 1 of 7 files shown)


jaypoulz:
Looks like it failed on an intermittent, so I restarted the test.

app/assets/javascripts/visualizations/highvis/bar.coffee
```diff
@@ -103,6 +103,30 @@ $ ->
...
+
+            errors.data = for fid in data.normalFields when fid in fieldSelection
+              barData = data.selector fid, gid
+              if barData.length == 1
+                []
+              else
+                mean = groupedData[fid][gid]
+                innerStd = barData.map((x) -> (x - mean) ** 2).reduce((x, y) -> x + y)
+                stdDev = Math.sqrt((1 / (barData.length - 1)) * innerStd)
+                if globals.configs.logY != true
```
jaypoulz: if !globals.configs.logY

app/assets/javascripts/visualizations/highvis/bar.coffee
```diff
@@ -103,6 +103,30 @@ $ ->
...
+                  [mean - stdDev, mean + stdDev]
+                else if mean > 0
+                  [mean, mean + stdDev]
+                else
+                  []
```
jaypoulz: Are negative means not useful? It seems to me that if mean >= 0 then [mean, mean + stdDev] else [mean, mean - stdDev]. Admittedly, this is just intuition, and not understanding. I would appreciate it if you would clarify this for me.

asalani93: Means that are negative on a logarithmic y axis are useless enough that they break the vis when you render the error bars over them.

`[mean, mean + stdDev]` renders an error bar from `mean` to `mean + stdDev`.  If `mean <= 0`, this throws an exception because some part of the error bar exists at or below zero.  Negative bars just don't show up on log y-axis either, so it'd be weird if I drew an error bar over a bar you can't see.

jaypoulz: Ah I see. This is specific to the log y axis. Y'okay.

PR #2166, opened 2015-06-04
Image hashing take two and code coverage increase

asalani93:
Addresses #2110 again.

New features:
* migration actually runs on live
* can no longer upload duplicate PNGs
* 'date modified' is removed from PNGs
* summernote now works correctly when you repeatedly press save or upload a duplicate media object

I also updated selenium because it wasn't working with the new versions of Firefox.

Increased code coverage on some files as well.  Also fixed a "bug" in the fields controller as part of this.  It's a two line fix so it's OK going in here.  It's not really even a bug either, the problem was that I'd convert literally every possible set of field restrictions to strings, and then check to see if they were strings.  This allowed for really weird field restrictions that would never be encountered in normal use of the website.  The only possible way to even hit this is by thoroughly abusing the API.


app/controllers/media_objects_controller.rb
```diff
@@ -101,15 +101,10 @@ def saveMedia
 
     # Rotate based on EXIF data, then strip it out.
```
bdonald25: This comment is no longer valid.

jaypoulz:
You're failing because of test coverage.

asalani93:
Yes.

db/migrate/20150514161235_add_hash_to_media_object.rb
```diff
@@ -0,0 +1,42 @@
+class AddHashToMediaObject < ActiveRecord::Migration
+  def isFeatured?(mo)
```
bdonald25: You are no longer deleting things so this function is not needed.

bdonald25:
No need for spinning earth gif

PR #2167, opened 2015-06-04
button to select bin size

bdonald25:
#2168
Adds a button to select bin size. 
Previously you would have to press the 'enter' key but this was confusing people at the boat house.


jaypoulz:
I claim this one since it's under vis

app/assets/javascripts/templates/visualizations/controls/histogram-tools.hbs
```diff
@@ -8,5 +8,6 @@
   <div>
     <label for="bin-size">Manual:</label>
     <input id="bin-size" class="form-control" value="{{binSize}}">
+    <button id="set-bin-size" class="btn btn-default">Set Bin Size</button>
```
jaypoulz: Multiple things with this line ->
1. in order to be compatible with Jim's work, it needs a log id. See the save hbs for examples.
2. the id should be 'set-bin-size-btn', as that is the precedent I've set in other tools.
3. Why is there no code associated with this button? Doesn't it's click need to be adjusted to the code that deals with hitting enter on input?

bdonald25:
@jaypoulz I added the button, clicked on it, and it worked as intended before I even looked at the code. :+1:

jaypoulz:
I've just tested this, and was terrified to find that it worked. I looked up the documentation for the 'change' method that is being bound. According to the jquery documentation,

```For select boxes, checkboxes, and radio buttons, the event is fired immediately when the user makes a selection with the mouse, but for the other element types the event is deferred until the element loses focus.```

Sooooo, it turns out that you can actually click anywhere that would cause the input box to lose focus and get the same behavior. That is why you're button is magically working. Still, this is concerning. I'm going to pull request a branch into this patch that refactors this the way I want it.

**Comment by jaypoulz** [comment]
Also, don't forget to add a ``data-log-id="set-bin-size"`` to the button handlebars file

File: app/assets/javascripts/templates/visualizations/controls/histogram-tools.hbs [hunk 2]
```diff
@@ -8,5 +8,6 @@
   <div>
     <label for="bin-size">Manual:</label>
     <input id="bin-size" class="form-control" value="{{binSize}}">
+    <button id="set-bin-size-btn" data-log-id="set-bin-size"class="btn btn-default">Set Bin Size</button>
```
jaypoulz: you need a space before class

Issue #2169, opened 2015-06-04
Icons on the bottom of the project boxes overflow.


bdonald25:
**General description:** Icons on the bottom of project boxes overflow.

**live/dev/localhost:** live
**iSENSE Version:** v7.1.1
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** X11
**Browser/Version:** Chrome 43.0.2357.81

**Steps to Reproduce:** On the project index page.

![overflow](https://cloud.githubusercontent.com/assets/4121924/7992248/1082f12c-0acc-11e5-8db5-6fb9fa591035.png)

393

bdonald25:
Testers: Try re-sizing the width of your browser and verify that the icons no longer overflow like in the image above.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).
Doesn't work for 3 different 5-digit numbers

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2170, opened 2015-06-04
Fixes overflow on the project boxes.

bdonald25:
#2169
The icons on the project boxes no longer overflow.


asalani93:
Looks good.  I tested with numbers that far exceed what iSENSE will have in the next year, and it still looks fine.

Issue #2171, opened 2015-06-04
Data set search button cuts off on mobile.


bdonald25:
**General description:** Search data set button cut off on mobile.

**live/dev/localhost:** live
**iSENSE Version:** v7.1.1
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** X11
**Browser/Version:** Chrome 43.0.2357.81

**Steps to Reproduce:** On a Project's page looking at it's data sets.

393

bdonald25:
Testers: Re-size the width of your browser to verify that the search button for searching data sets on a project page no longer cuts off on small screens.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
FAILED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2172, opened 2015-06-04
data set search button is no longer cut off on mobile

bdonald25:
#2171


asalani93:
You should do what I did for the project and vis search pages.

Issue #2157, opened 2015-05-29
Slickgrid text field values not persisting across edits


asalani93:
**General description:** Selecting a value from a text field in either the data set edit table or manual entry table makes that value vanish.

**live/dev/localhost:** dev
**iSENSE Version:** 7.1
**Logged in (Y or N):** Y
**Admin (Y or N):** Y/N

**OS:** Ubuntu 14.04 x64
**Browser/Version:** Chrome 42 x64

**Steps to Reproduce:**
* Edit a data set with text fields
* Highlight a text field
* Observe as the text field vanishes

**Associated Image(s):**

bdonald25:
Testers: Edit data -> click on a text field -> verify data did not disappear

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2173, opened 2015-06-05
Fixing text fields

asalani93:
Addresses #2157.  Also fixed some bugs with regards to how the caret moved around in the timestamp and text with restrictions fields.

I'm not sure why some of the image hashing pull request got thrown in here.  That's weird.  Ignore that.  Ignore the first four pull requests when code reviewing I guess.


bdonald25:
@asalani93 Failed because it needs more test coverage.

Issue #413, opened 2013-07-26
Data filtering


mmcguinn:
Ability to "hide" data not currently visible in the vis. This will be for cross-vis usage.
Will enable the functionality described in #1090.

Finally, the UI interface for this might be similar to regressions, and might be titled "data filters."

bdonald25:
Testers: Try to break filtering. This will require thorough testing of all the visualizations.

jaypoulz:
omg it's really going to be a thing. I think I'm tearing up. :sob:

asalani93:
PASSED testing on Chrome 45.0.2454.101 (linux).
minor issues:
* filtering by text fields puts NaN in the top part with the list of filters (filtering still works though)
* "contains" is case-insensitive in table and case-sensitive in filtering

tylerpuleo:
PASSED testing on Chrome 45.0.2454.101 (linux).

PR #2181, opened 2015-06-10
Data clipping built on top of vis 2.0

jaypoulz:
Featuring a new user interface that is theoretically user friendly

Takes care of #413


jaypoulz:
@jdalphon wanna take this for a test drive?

jaypoulz:
@bdonald25 The scrolling issue you were having has been addressed

jaypoulz:
bump

bdonald25:
@jaypoulz Turning filtering on and off on scatter still makes the data disappear.

bdonald25:
@jaypoulz Filtering on map requires you to go to another vis and then go back to map.

asalani93:
Just tried it out again - the only problem I had was what Bobby mentioned - filtering on the map requires you to reload the map for the change to take affect.  Couldn't reproduce the bug with scatter.

There was also a point where all of my filters disappeared, but I'm pretty sure that was my fault and I just forgot what I did to make that happen.

It seems pretty sturdy and I couldn't find any ways to break it.

jaypoulz:
I can't reproduce the scatter issue. Is there a console error that comes with this? Or could you give me specific step by step instructions including platform and browser?

bdonald25:
No console error. I am looking at the data from the bison project (1158) on the latest version of chrome on Ubuntu 15.04.
Go to scatter -> zoom in on data -> click the switch to turn on filtering -> watch scatter turn into a blank graph

bdonald25:
It is not project specific though. It happens on every project.

jaypoulz:
Alright, so the latest version fixes the map bug, and disables animations on scatter so that is doesn't look like it reloads even though it does.

jaypoulz:
This is ready for a last round of review. I've tweaked the title bar in hopes of making it more mobile friendly/collapsible. For this round of testing, it might be good to mess with browser size and see what happens.

bdonald25:
The hamburger button is not big enough.

asalani93:
I concur with Bobby.  The hamburger is clearly not large enough.  Right now, it's sitting at around 1/3 lbs.  It needs to be at least 2/3 lbs or more.

:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger: 
:hamburger:

bdonald25:
Zooming in on scatter does not work for me. If I drag a perfect square it does nothing, if I make a rectangle it kind of zooms in but not correctly.

asalani93:
If the same field is on the x and y axis in scatter, filtering from that does weird things:
filtering-1

jaypoulz:
@bdonald25 I can't reproduce the scatter zoom problem, but I had that exact problem before I updated my data base. I even had the zoom issue on my master branch. Honestly, I'm not sure what is going on there. This version seems to work for me with filtering on or off. I checked both projects and saved vises. If we can reproduce it, I'll patch it, but right now it seems clean.

jaypoulz:
I think we're ready to pull here unless someone wants to make style comments.

asalani93:
hamburgre

jaypoulz:
@asalani93 such burger, much meat. wow. :meat_on_bone: :hamburger:

asalani93:
I'm not joking.  Make the hamburger smaller.

jaypoulz:
@asalani93 Unless you have a suggestion on how the top bar should look, you're going to have to be more constructive.

bdonald25:
I think it looked better before. It still overflows just as easily on phones but you now sacrifice screen space.

asalani93:
Jim suggested to make the button smaller and look like a part of the bar.  I like that idea.  Also, instead of a hamburger, maybe an arrow might be more indicative of what it does.

jaypoulz:
@bdonald25 I'm inclined to disagree - but I have reasons:
1. The old format would prevent you from having access to project link-back for even moderate length titles. The overflow is less of an issue here, so much as easy access to the functionality that is supposed to be there.
2. The reason I changed the top bar was to deal with the data set title edit button + the refresh button. The problem with having a filtering div below the top bar is that it pushes into the tab listing for the vises and causes strange layouts in almost any dimensions below that of a desktop screen.

@asalani93 If you could draw me an example of what you are looking for, I'll try to implement it. Otherwise I can only work within my own mental context.

jdalphon:
Looks awesome but I have at least 2 more requests for you.

1) The menu (hamburger) is too big. Please try to make it a bit smaller. It would be nice if it could look like it was more part of the menu, but that can be fixed later. 
 
2) When filters are cleared, I think it needs to clear the top bar of the table. This just lead to a bunch of confusion in the lab. 

Thanks

asalani93:
Here's something I just threw together:
suggest

jaypoulz:
@jdalphon
1. I'll see what I can do.
2. That's a reasonable fix. I came across this when testing, but I guess being the developer makes it easier to understand what is going on.

@asalani93 It's funny because the first design I came up with for the tools expansion was exactly what you drew. I implemented it, and decided that the embedded hamburger button looked nicer.

I'm going to try to take your suggestions about how to implement the tools UI, but in all likelyhood I'll end up doing something different. Check again (probably Thursday) for a new iteration.

jdalphon:
@jaypoulz  thanks

bdonald25:
opened pull request #2230

Issue #2004, opened 2015-02-13
Duplicate colorpickers


asalani93:
**General description:** Clicking on the colorpicker buttons in a certain way makes two colorpickers.  It's not that bad because only two at max can stay open, and when you do anything, at least one will go away.

**live/dev/localhost:** dev
**iSENSE Version:** 6.1
**Logged in (Y or N):** Y/N
**Admin (Y or N):** Y/N

**OS:** Ubuntu 14.04.1 x64
**Browser/Version:** Chrome 39.0 x64

**Steps to Reproduce:**
* Open a vis w/ more than one dataset
* Switch to timeline/scatter/whatever
* Open colorpicker
* Change S or V (not hue, basically)
* Before you mouse up, move the mouse outside of the colorpicker pane
* Release the mouse
* Click on a different colorpicker icon
* Observe as there are now two open colorpickers

**Associated Image(s):**
![selection_005](https://cloud.githubusercontent.com/assets/7584534/6191415/4479e0be-b377-11e4-9f50-4cf64c8dcd81.png)

bdonald25:
Testers: Follow the steps above and make sure you do not end up with two color pickers.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2189, opened 2015-06-17
Fixing duplicate colorpickers

asalani93:
Addresses #2004.

Companion pull request for https://github.com/isenseDev/better-colorpicker/pull/1.  Don't do anything with this until that is merged.

This gets rid of the old colorpicker.  It was a good first try, but it had a couple of problems.  Primarily, it would created one instance of itself in the DOM per group.  This is OK on most projects, but was still bad practice.  It also allowed more than one color picker to exist on screen at a time.

This should probably be tested locally before being merged.


bdonald25:
Ready to merge once the color mixer pull request is merged.

Issue #2188, opened 2015-06-17
Vis not displaying correctly for some data sets


asalani93:
**General description:**
Look here:
http://isenseproject.org/projects/891/data_sets/8083
I can't visualize the data at all.  No idea what's going on.

**live/dev/localhost:** all
**iSENSE Version:** 7.1.1
**Logged in (Y or N):** Y/N
**Admin (Y or N):** Y/N

**OS:** Ubuntu 14.04 x64
**Browser/Version:** Chrome 43 x64

**Steps to Reproduce:**
* Go here: http://isenseproject.org/projects/891/data_sets/8083
* Observe that you can't observe the vis

**Associated Image(s):**
![datapocalypse-1](https://cloud.githubusercontent.com/assets/7584534/8208144/ef7fbe8a-14d3-11e5-981c-c7718499e49c.png)
![datapocalypse-2](https://cloud.githubusercontent.com/assets/7584534/8208146/ef81d1d4-14d3-11e5-9492-44b69e5cbed7.png)
![datapocalypse-3](https://cloud.githubusercontent.com/assets/7584534/8208145/ef80baba-14d3-11e5-97dd-9f76079205e3.png)

jaypoulz:
One of the saved regressions is malformed for this data, which is causing a nullity error in the creation of the regression controls. If this is the only data set with this error, it may just be worth going in from the db console and dropping the savedRegression data altogether.

jaypoulz:
The only savedRegression for this data set in an object that has:
```
bounds {x bounds,  y bounds}
fieldIndices {x index, y index}
fieldNames {x field name, y field name}
regressionId = "" <-- this is strange
series = {highcharts configs like color and line type}
type = 0 // implies linear regression
typeCount = 0 // for old naming conventions
```

This doesn't not meet the criterion for the new format. I suspect either a migration has a bug, or one failed to run on live. Either way, the patch I submitted treats the symptom, but not the cause.

PR #2191, opened 2015-06-18
Added nullity checkers to saved regressions

jaypoulz:
Towards #2188 

I don't know that this will allow for the vis to load completely, but it will pass these malformed objects.


asalani93:
Still doesn't work.  It breaks on every available vis, not just scatter.  If this is a one time thing caused by a bad migration, it probably would just be better to delete it to be honest.

jaypoulz:
This should still be pulled, because it makes iSENSE less prone to displaying nothing. I agree about the current broken things on live though.

Issue #2187, opened 2015-06-16
Can't sort photos


smichalka:
**General description:**
The Water SCIENCE program would like to be able to have the same "Group by" option for the Photos to allow them to look at a particular team's photos.

**live/dev/localhost:**
**iSENSE Version:**
**Logged in (Y or N):**
**Admin (Y or N):**

**OS:**
**Browser/Version:**

**Steps to Reproduce:**

**Associated Image(s):**

bdonald25:
Testers: Upload images to different data sets. When you visualize them together you should be able to group by data set. Make sure this works as expected.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

smichalka:
Awesome.  This looks great.  Thanks so much everyone (especially Bobby, I
assume)!

jaypoulz:
@smichalka Actually Anthony did the work for this one :smile: Go go @asalani93

PR #2192, opened 2015-06-18
Adds grouping to photo vis

asalani93:
Addresses #2187.

You can now group and filter out photos based on what data set they are in.

### DIFF
--- a/app/assets/javascripts/templates/visualizations/photo/pic.hbs
+++ b/app/assets/javascripts/templates/visualizations/photo/pic.hbs
@@ -1,5 +1,5 @@
 <!-- Handlebars template for the photo box in the photo vis -->
-<div class="col-xs-6 col-lg-3 pic-box">
+<div class="col-xs-6 col-md-4 col-lg-3 pic-box">
   <div id="{{p_id}}" class="item item-block clickable">
     <div class="item-image item-photo-image"
          style="background-image:url({{tn_src}})">
--- a/app/assets/javascripts/templates/visualizations/photo/pic.hbs
+++ b/app/assets/javascripts/templates/visualizations/photo/pic.hbs
@@ -1,5 +1,5 @@
 <!-- Handlebars template for the photo box in the photo vis -->
-<div class="col-xs-6 col-lg-3">
+<div class="col-xs-6 col-lg-3 pic-box">
   <div id="{{p_id}}" class="item item-block clickable">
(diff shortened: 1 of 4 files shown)


jaypoulz:
Out html checker is down again. I though I'd fixed this, but apparently not.

bdonald25:
You should hide groups that don't contain any photos.

app/assets/javascripts/templates/visualizations/photo/pic.hbs
```diff
@@ -1,5 +1,5 @@
 <!-- Handlebars template for the photo box in the photo vis -->
-<div class="col-xs-6 col-lg-3">
+<div class="col-xs-6 col-lg-3 pic-box">
```
bdonald25: You should add col-md-4.

bdonald25:
Testers: Upload images to different data sets. When you visualize them together you should be able to group by data set. Make sure this works as expected.

app/assets/javascripts/visualizations/highvis/photos.coffee
```diff
@@ -52,39 +46,69 @@ $ ->
...
-                $(canvas).append lbTemp(ctx)
-
-                $('#target-img').modal
-                  keyboard: true
-                $('#target-img').on 'hidden.bs.modal', ->
-                  $('#target-img').remove()
-
-              id++
+        for group in selectedGroups
+          if groupedMedia[group[1]].length == 0
```
jaypoulz: What if groupedMedia[group[1]] is undefined? (this is occurring in my localhost)

app/assets/javascripts/visualizations/highvis/photos.coffee
```diff
@@ -52,39 +46,69 @@ $ ->
...
+            continue
+
+          groupContext =
+            g_id: group[0]
+            group_label: group[1]
+          groupObj = $(grpTemp(groupContext))
+
+          # put the photos in their correct groups
+          for photo in groupedMedia[group[1]]
```
jaypoulz: Again, groupedMedia[group[1]] may not exist

jaypoulz:
We should verify that projects with pre-existing photos are not broken, as is the case on my local host.

jaypoulz:
Also, let's try to be more conscious of nullity checks in our future vis patches - it's better for vis to skip draw a small section than to abort mid-draw

Issue #2178, opened 2015-06-09
bargraph view should default to mean


fgmart:
It's the most common use of bar.

right now the default of total is confusing.

probably mean should first in the list too.

bdonald25:
Testers: Make sure bar graph defaults to mean and that mean is the first analysis type.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2193, opened 2015-06-19
Sets mean to be default analysis mode for bar chart

asalani93:
Addresses #2178.

If there's anything else that needs to be changed, let me know.


bdonald25:
This sets the mean as default but it does not set it to be first in the list of analysis types.

Issue #1734, opened 2014-08-13
Add option in user's page for timezone setting


AlanRosenthal:
**General description:**
User timezone setting would be awesome. Timestamp fields would be much easier for teachers to use.

**live/dev/localhost:** N/a
**iSENSE Version:** N/a
**Logged in (Y or N):** N/a
**Admin (Y or N):** N/a

**OS:** N/a
**Browser/Version:** N/a

**Steps to Reproduce:** N/a

**Associated Image(s):** None

bdonald25:
Testers: Make sure time stamps on iSENSE are being displayed correctly using your timezone.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
PASSED testing on Safari 8.0 (mac).

PR #2198, opened 2015-06-22
Display time in clients timezone instead of gmt.

bdonald25:
#1734  This gets the clients timezone and converts to that timezone.


asalani93:
This looks good.  I'll have to make some changes to manual entry and the date time picker to support timezones correctly, however.

Issue #2203, opened 2015-06-24
Detailed tool tips on timeline don't show


asalani93:
**General description:**
It's like some sort of bug or something.

**live/dev/localhost:** live
**iSENSE Version:** 7.2
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** Ubuntu 14.04 x64
**Browser/Version:** Chrome 43 x64

**Steps to Reproduce:**
* Click "Detailed Tooltips"
* Mouse over a timeline
* Observe the lack of tooltips

**Associated Image(s):**

bdonald25:
Testers: Verify that detailed tool tips work on scatter.

qianacurcuru:
PASSED testing on Chrome 43.0.2357.132 (mac).

bdonald25:
PASSED testing on Chrome 43.0.2357.132 (linux).

asalani93:
PASSED testing on Chrome 43.0.2357.130 (linux).

PR #2205, opened 2015-06-24
fix detailed tool tips for timeline

bdonald25:
#2203 Fixes detailed tool tips.
My commit says scatter but it is really for timeline.


jaypoulz:
I haven't verified it, but this looks like the correct fix to me.

Issue #2067, opened 2015-03-13
Contributor keys are inconsistent


stowellm:
**General description:** Contributor names are not used.  At all.  Anywhere.  They seem entirely useless when using a contributor key through the site.  The name that shows up next to the data set name in the data set list has become the label for the key - I feel like it *should* be the contributor name.  

Also, uploading through the API with a contributor key produces inconsistent results.  Not only does a gravatar icon show up instead of the key icon, but *now* the contributor name shows.  

**live/dev/localhost:** dev
**iSENSE Version:** 6.3.something
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu 14.04 LTS
**Browser/Version:** Chrome 41.something

**Steps to Reproduce:**
Read the description.

**Associated Image(s):**

bdonald25:
Testers: Give a project a contributor key. Upload data using that key and make sure that the contributor name shows up as expected.
 Also, try uploading data with one of the android applications using a contributor key.

asalani93:
PASSED testing on Chrome 43.0.2357.81 (linux).

bdonald25:
PASSED testing on Chrome 43.0.2357.125 (linux).

qianacurcuru:
PASSED testing on Chrome 43.0.2357.130 (mac).

KatherineBrunelle:
FAILED testing on Safari 8.0 (mac).

jaypoulz:
There is no commit associated with this patch... was it ever really in testing?

bdonald25:
Anthony and I thought it was fixed a long time ago so we just threw it in
testing to find out. Turns out it was not fixed.

jaypoulz:
Hahaha ok :+1: :smile_cat:

bdonald25:
Testers: Give a project a contributor key. Upload data using that key and make sure that the contributor name shows up as expected.
Also, try uploading data with one of the android applications using a contributor key.

qianacurcuru:
FAILED testing on Chrome 43.0.2357.132 (mac).

bdonald25:
PASSED testing on Chrome 43.0.2357.132 (linux).

asalani93:
PASSED testing on Chrome 43.0.2357.130 (linux).

PR #2210, opened 2015-06-25
Contributor names now appear where they should

asalani93:
Addresses #2067.

### DIFF
--- a/app/controllers/data_sets_controller.rb
+++ b/app/controllers/data_sets_controller.rb
@@ -72,7 +72,10 @@ class DataSetsController < ApplicationController
       return
     end
 
-    if @cur_user.nil?
+    if !session[:contributor_name].nil?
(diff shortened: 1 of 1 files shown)


app/controllers/data_sets_controller.rb
```diff
@@ -72,7 +72,10 @@ def create
       return
     end
 
-    if @cur_user.nil?
+    if !session[:contributor_name].nil?
```
bdonald25: if session[:contributor_name]?

asalani93: apparently that is not valid ruby

bdonald25: unless session[:contributor_name].nil?

asalani93: That makes rubocop throw up if you have an elsif or else following it.

Issue #2202, opened 2015-06-24
Can crash the vis index the same way as the project index


asalani93:
**General description:**
Same as #2196, but for the vis index page this time.

**live/dev/localhost:** live
**iSENSE Version:** 7.2
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu 14.04 x64
**Browser/Version:** Chrome 43 x64

**Steps to Reproduce:**
* Same as before

**Associated Image(s):**

bdonald25:
Testers: Follow the steps on #2196 but for saved vises.

qianacurcuru:
PASSED testing on Chrome 43.0.2357.132 (mac).

bdonald25:
PASSED testing on Chrome 43.0.2357.132 (linux).

asalani93:
PASSED testing on Chrome 43.0.2357.130 (linux).

PR #2217, opened 2015-07-06
fixing viz index bug.

jdalphon:
Wrong way to check if a attribute is set.


bdonald25:
Fixes the symptoms of #2202

Issue #2142, opened 2015-05-21
Spider vis


bdonald25:
**General description:** The labels make pi chart cluttered on mobile. It would be more useful to have a bigger pie chart. You can get the same info by pressing on each section of the chart.

**live/dev/localhost:** live
**iSENSE Version:** v7.0
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** Macintosh
**Browser/Version:** Chrome 42.0.2311.152

**Steps to Reproduce:** Go to pie chart on a project with a lot of data sets on mobile.

![screenshot_2015-05-20-23-32-55](https://cloud.githubusercontent.com/assets/4121924/7741109/0fd281d6-ff49-11e4-862f-b1cf10032f9c.png)

393

jaypoulz:
I don't think there is much we can do for this. We simple don't have the screen real-estate. Maybe we could tweak it so that labels are in a legend on small screens?

bdonald25:
Testers: Make sure pie chart looks sane on all screen sizes.

Issue #2141, opened 2015-05-21
Field vis


bdonald25:
**General description:** Bar chart on mobile is not visible when there are a lot of datasets.

**live/dev/localhost:** live
**iSENSE Version:** v7.0
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** Macintosh
**Browser/Version:** Chrome 42.0.2311.152

**Steps to Reproduce:** Go to bar chart on a project with a lot of data sets.

![screenshot_2015-05-20-23-29-23](https://cloud.githubusercontent.com/assets/4121924/7741105/fd06bd06-ff48-11e4-8dd6-b86a8ca92231.png)

393

jaypoulz:
Ok so scratch the legend idea for pie charts haha. Once again real estate is a major drawback here.

asalani93:
Couldn't legends and labels be hidden on mobile?  For example, you could get a label for pie chart by clicking the slice of the pie you want to investigate.  Same for bar chart.  I'm not too familiar with High Charts though, so I don't know if this feasible or not.

bdonald25:
It is very easy to turn off the legend with high charts. It just requires a boolean to be set to false.

fgmart:
this isn't just a mobile issue (see #2176). 

Perhaps we can put the key in a scrolling window (at least on desktop)? I guess that's terrible, what are better ideas?

bdonald25:
I think you should hide them altogether if you have that many. There are so many colors that make it hard to pick out which bar is which anyways. In this image there are at least 11 green bars. You would still be able to see which data set a bar is by clicking on it.

jaypoulz:
I agree with Bobby. That also makes this a fairly straightforward fix. Probably shouldn't be assigned to Jake tho.

fgmart:
OK, how to operationalize "too many" ?

On Mon, Jun 8, 2015 at 5:11 PM, Jeremy Poulin <notifications@github.com>
wrote:

> I agree with Bobby. That also makes this a fairly straightforward fix.
> Probably shouldn't be assigned to Jake tho.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/isenseDev/rSENSE/issues/2141#issuecomment-110141675>.
>



-- 

Work Ready Life Ready World Ready

Director, Engaging Computing Group <http://www.cs.uml.edu/ecg>
Associate Professor, Computer Science <http://blog.uml.edu/cs>
[...]

PR #2218, opened 2015-07-07
Fix spider vis and field vis

bdonald25:
#2142 #2141 
This pull request will hide the keys for pie chart and bar chart when they are overwhelming.
I plan on making a second pull request that will add a button to the tools that will force the keys to show.


app/assets/javascripts/visualizations/highvis/bar.coffee
```diff
@@ -75,6 +75,8 @@ $ ->
              }]
           xAxis:
             type: 'category'
+          legend:
+            enabled: if $(window).width() > 700 && data.groups.length < 20 then true else false
```
jaypoulz: Use `and` not `&&`
Also the then true else false implies this would be the same line:
``enabled: $(window).width() > 700 and data.groups.length < 20``

app/assets/javascripts/visualizations/highvis/pie.coffee
```diff
@@ -96,7 +96,8 @@ $ ->
               allowPointSelect: true
               cursor: 'pointer'
               dataLabels:
-                enabled: true
+                floating: false
+                enabled: if $(window).width() > 700 && data.groups.length < 50 then true else false
```
jaypoulz: same comments as above

stowellm:
This PR name caught my attention... wat is "spider vis"?

bdonald25:
@stowellm If you look at pie chart on mobile when there are a lot of groups it looks like a spider. 

The pie is really small and the labels look like legs coming off of it.

jaypoulz:
I think I speak for myself and @stowellm when I ask: can we make a :goat: vis?

app/assets/javascripts/visualizations/highvis/pie.coffee
```diff
@@ -93,10 +92,12 @@ $ ->
             useHTML: true
           plotOptions:
             pie:
+              showInLegend: ($(window).width() < 1000 && data.groups.length < 30) && !($(window).width() < 400)
               allowPointSelect: true
               cursor: 'pointer'
               dataLabels:
-                enabled: true
+                floating: false
+                enabled: ($(window).width() > 1000)
```
asalani93: screen size of 1000px wide will show neither

Issue #2234, opened 2015-08-12
CORS doesn't work correctly


asalani93:
**General description:** Cross-Origin resource sharing doesn't work correctly when the server sends back responses with statuses other than 200 or 302.  This OK for most applications that use the API, but anything that uses Javascript and runs in the browser will get a CORS-related error and they won't be able to see how their call failed.

**live/dev/localhost:** live/dev/localhost
**iSENSE Version:** 7.2.2
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** Ubuntu 14.04.1 x64
**Browser/Version:** Chrome 43 x64

**Steps to Reproduce:**
* make an invalid API request (try GET on http://isenseproject.org/api/v1/projects/99999)
* observe that you get a CORS exception and the status on the response is 0, not 404

**Associated Image(s):**

PR #2235, opened 2015-08-12
Changes API to correctly handle CORS

asalani93:
Addresses #2234.

Now using Rack CORS to allow CORS instead of whatever we had before.


asalani93:
TESTERS (OR OTHER):

Open the JS console on a website that isn't iSENSE and uses jQuery.  Copy and paste this into the console:
`$.ajax({
  async: false,
  dataType: 'JSON',
  type: 'GET',
  url: 'http://rsense-dev.cs.uml.edu/api/v1/projects/500000000'
});`

PR #2240, opened 2015-08-17
Fix Travis

bdonald25:
This will move us off of legacy Travis. It should fix all our texting issues and allow tests to run faster.


bdonald25:
This is failing because Travis has not white listed libreoffice so it fails to install. Travis has an open pull request to fix this but it has merge conflicts. Once that pull request makes it's way into travis, this one will pass testing.

Issue #2236, opened 2015-08-12
saved vis's don't have nice icons anymore


fgmart:
hey, sometime after June 23, saved vis's no longer get mini-screens of the vis anymore.

i know that we never did this for maps, but we used to do this for bar, scatter, histo, etc.

See below... what happened? 

![screen shot 2015-08-12 at 2 16 50 pm](https://cloud.githubusercontent.com/assets/1038856/9232660/e941bac8-40fc-11e5-9725-d3fd1cbc1505.png)

bdonald25:
Testers: Save a vis and make sure the thumbnails are mini-images of the vis. (does not work with maps)

pat-son:
PASSED testing on Chrome 45.0.2454.99 (mac).

PR #2244, opened 2015-08-18
Saved visualizations get thumbnails again

asalani93:
Addresses #2236.

Not sure how this worked until recently.

### DIFF
--- a/app/controllers/visualizations_controller.rb
+++ b/app/controllers/visualizations_controller.rb
@@ -106,7 +106,6 @@ class VisualizationsController < ApplicationController
 
     if params[:visualization].try(:[], :svg)
       begin
-        puts "\n\nAAAA\n\n"
         mo = MediaObject.new
         mo.media_type = 'image'
         mo.name = 'image.png'
--- a/app/controllers/visualizations_controller.rb
+++ b/app/controllers/visualizations_controller.rb
@@ -102,15 +102,18 @@ class VisualizationsController < ApplicationController
     end
 
     # Try to make a thumbnail
-    mo = MediaObject.new
-    mo.media_type = 'image'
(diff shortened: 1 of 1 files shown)


app/controllers/visualizations_controller.rb
```diff
@@ -102,15 +102,18 @@ def create
...
-    mo.media_type = 'image'
-    mo.name = 'image.png'
-    mo.file = 'image.png'
-    mo.user_id = @cur_user.id
-    mo.check_store!
+    mo = nil
 
     if params[:visualization].try(:[], :svg)
       begin
+        puts "\n\nAAAA\n\n"
```
bdonald25: puts

app/controllers/visualizations_controller.rb
```diff
@@ -120,8 +123,9 @@ def create
         end
 
         mo.add_tn
-
       rescue MiniMagick::Invalid => err
+        mo = nil
+        puts "\n\nBBBB\n\n"
```
bdonald25: puts

app/controllers/visualizations_controller.rb
```diff
@@ -102,15 +102,17 @@ def create
...
       begin
+        mo = MediaObject.new
+        mo.media_type = 'image'
+        mo.name = 'image.png'
+        mo.file = 'image.png'
+        mo.user_id = @cur_user.id
+        mo.check_store!
+
         image = MiniMagick::Image.read(params[:visualization][:svg], '.svg')
         image.format 'png'
```
bdonald25: @asalani93 One thing I noticed about saving a vis that was shown by the imagemagick issue is that if it ever fails here then the vis page will hang trying to save.

Issue #2237, opened 2015-08-14
news page only provides one screen of results


fgmart:
The news page only displays the most recent 10 items. It does not provide a way of paging to earlier items.

Please fix.

fgmart:
this bug also exists for tutorials:

https://isenseproject.org/tutorials

shows 8, but there are more than this.

bdonald25:
Testers: Make sure news and tutorial index pages have more than one page of results.

pat-son:
PASSED testing on Chrome 45.0.2454.99 (mac).

PR #2246, opened 2015-09-02
pages for tutorials

bdonald25:
#2237


asalani93:
Failed once with
`Error:
ContribKeyWithDataSetTest#test_website_and_API:
NoMethodError: undefined method '[]' for nil:NilClass
    /home/asalani/Documents/ecg/rSENSE/test/integration/contrib_key_with_data_set_test.rb:53:in 'block in <class:ContribKeyWithDataSetTest>'`

Then passed twice.  I tried again with the same seed and can't replicate it, so I guess it passes.

PR #2267, opened 2015-10-08
Re-order Fields

bdonald25:


asalani93:
Fails the four tests that are fixed in another pull request, but that's OK.  Also fails 'test_website_and_API' intermittently, but master is currently failing that for me so it's not the fault of your code.

Tested the cases that failed before and they work just fine.  Apparently the way you wrote this, you can create fields with really weird indices (like -20, 1.5, "dog", etc...) via the API but the server handles this fine?  So, it's not a problem.  Fun fact: an index of "dog" puts that field at the top of the list of otherwise numerically indexed fields.

Also, code looks good (basically the same as before).

PR #2269, opened 2015-10-09
fixes tests

bdonald25:
Fixes four tests that are failing. 

The message they get is something along the lines of "expected 'over 239 years' but got 'almost 240 years'"


asalani93:
Passed local testing.

Issue #2262, opened 2015-09-28
Map visualization is enabled when there is no location data.


bdonald25:
**General description:** Map visualization is enabled when there is no location data. 

**live/dev/localhost:** live
**iSENSE Version:** v7.3.0
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** X11
**Browser/Version:** Chrome 45.0.2454.101

**Steps to Reproduce:** 

393

asalani93:
PASSED testing on Chrome 45.0.2454.101 (Ubuntu 15.04 x64).
==yes==

bdonald25:
PASSED testing on Chrome 46.0.2490.80 (linux).

bdonald25:
PASSED testing on Chrome 46.0.2490.80 (linux).

tylerpuleo:
FAILED testing on Chrome 45.0.2454.101 (Mint).
Datasets uploaded before the location fields were added still enable the map.

Steps to reproduce: Create a project with number and or text fields. Upload a dataset using these fields. Add location fields. Look at old datasets and map will be enabled.

# NO

bdonald25:
@pat-son 
We are still going to release the current fix you made because it does fix the problem for every case except data sets that already existed. I will leave this issue open for you to fix that last issue.

bdonald25:
Testers: Create a project that contains location fields. Upload data that contains data for other fields but does not contain location data. Make sure the map visualization is not enabled.

PR #2278, opened 2015-10-22
Map vis is disabled when there isn't any location data

pat-son:
#2262 
Fixed Map Vis so that it is correctly disabled when no location data is present.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -64,6 +64,7 @@ ActiveRecord::Schema.define(version: 20150519161334) do
     t.integer  "news_id"
     t.string   "store_key"
     t.string   "file"
+    t.string   "md5"
   end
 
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -158,7 +151,6 @@
     t.datetime "featured_at"
     t.text     "summary"
     t.integer  "thumb_id"
-    t.boolean  "should_log",  default: false
   end
 
-end
\ No newline at end of file
+end
```
asalani93: Don't commit these changes

asalani93:
All tests pass aside from the four that were fixed last week, which will be fixed upon merging.  Code looks really code, especially for your first pull request.  Good job on adding tests, too.

Issue #2279, opened 2015-10-23
Field renaming quirk


asalani93:
**General description:** Can't create a Timestamp, Latitude, or Longitude field and give it a name at the same time.  You must first create the field, save it, and then rename it.  Text and Number do not work like this.

**live/dev/localhost:** dev
**iSENSE Version:** 7.5
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu 15.04 x64
**Browser/Version:** Chrome 4x x64

**Steps to Reproduce:**
* create a project
* add a timestamp,  latitude, and longitude field.  do not save the fields yet
* give each a custom name (eg: timestamp => a, latitude => b, longitude => c)
* save the fields
* observe how the name didn't change
* try and change the names again
* observe how the names did change this time
* try the above for numbers and text, and observe how they work right the first time

**Associated Image(s):**

bdonald25:
Testers: While manually entering fields, change the names of Timestamp, Latitude, and Longitude fields to be something different and make sure they are actually saved with their new names.

asalani93:
PASSED testing on Chrome 45.0.2454.101 (linux).

pat-son:
PASSED testing on Chrome 46.0.2490.80 (Mac OSX 10.10.5).

JasonD94:
PASSED testing on Chrome 46.0.2490.80 (linux).

PR #2280, opened 2015-10-23
Set Lat, Lon, and Timestamp names correctly.

bdonald25:
#2279


asalani93:
rubocop for double quoted strings

PR #2281, opened 2015-10-23
Map vis no longer enabled for data sets without location data added before location fields

pat-son:
#2262


bdonald25:
Failed Rubocop.

pat-son:
Closed because [another request](https://github.com/isenseDev/rSENSE/pull/2357) includes these fixes.

Issue #2294, opened 2015-11-10
improve error message when Timeline isn't displayed


fgmart:
Presently it says this:

> Either a time field was not found or there were not enough data 
> Cannot display Timeline visualization

As a trivially easy improvement, I would like to revise to:

> Either Timestamp field or Number field is missing from project
> Cannot display Timeline visualization

Thanks.

bdonald25:
Testers: When manually adding fields to a project, make sure it is clearly documented what fields you need for timeline to work.

asalani93:
PASSED testing on Chrome 45.0.2454.101 (linux).
can still get the screen if you don't have data in all three data points

pat-son:
PASSED testing on Chrome 46.0.2490.80 (Mac OSX 10.10.5).

JasonD94:
PASSED testing on Chrome 46.0.2490.80 (linux).

Issue #2293, opened 2015-11-10
additional user prompt when setting up Timeline field


fgmart:
Sam and I both just lost 15+ mins ea trying to figure out why a teacher's project wouldn't come up with Timeline, even though she had done everything properly in terms of formatting dates.

The answer was: her CSV did not have any numeric data to display.  (ARGHHHHH!)

Let's please add the following (highlighed in bold, but don't add it in bold) to the help-text below Timeline when setting up fields:

> Data and time values in the form yyyy/mm/dd hh:mm:ss. Limit 1 per project. **Project must have at least one Number field.**

Thanks.

bdonald25:
Testers: Go to a visualization that timeline will be disabled for. Make sure the error message clearly explains why timeline is disabled. 

For timeline to be enabled you need a number field, timestamp, and at least 3 data points.

asalani93:
PASSED testing on Chrome 45.0.2454.101 (linux).

pat-son:
PASSED testing on Chrome 46.0.2490.80 (Mac OSX 10.10.5).

JasonD94:
PASSED testing on Chrome 46.0.2490.80 (linux).

PR #2296, opened 2015-11-10
Better timeline errors

bdonald25:
Addresses #2293 and #2294.

### DIFF
--- a/app/assets/javascripts/visualizations/highvis/disabledVis.coffee
+++ b/app/assets/javascripts/visualizations/highvis/disabledVis.coffee
@@ -40,7 +40,7 @@ $ ->
           <br>
           <br>
           Either a timestamp or number field is missing,
-          or there is not at least 3 data points.
+          or there are not at least 3 data points.
           <br>
           Cannot display Timeline visualization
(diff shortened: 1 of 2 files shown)


asalani93:
fails coffeelint

asalani93:
different coffeelint error

app/assets/javascripts/visualizations/highvis/disabledVis.coffee
```diff
@@ -39,7 +39,8 @@ $ ->
           <img src='#{window.icons["novis_timeline"]}'>
           <br>
           <br>
-          Either a time field was not found or there were not enough data
+          Either Timestamp field or Number field is missing from project or
+          there is not enough data. (min of 3 points required)
```
asalani93: This just sounds awkward.  How about:
"Either a timestamp or number field is missing, or there is not at least 3 data points."

fgmart: It would need to be, "there are not at least three data points".  ( Change
" is" to "are")

PR #2297, opened 2015-11-10
"Showing x data sets" -> x is now the correct number of data sets

bdonald25:
When you visualize 5 of the 10 data sets the visualization will now correctly say "Showing 5 Data Sets" instead of "Showing 10 Data Sets"


asalani93:
fails rubocop

Issue #2276, opened 2015-10-21
this project breaks the visualization - media box at top, no data appears


smichalka:
This is not good:
![image](https://cloud.githubusercontent.com/assets/5709961/10645521/266e6b36-77fc-11e5-99b9-a32869bb95b7.png)

Project number 891.  However, when I cloned the project with data, everything looks fine (project 1450).

bdonald25:
Testers: Make sure http://rsense-dev.cs.uml.edu/projects/891/data_sets/8083 is not broken.

asalani93:
PASSED testing on Chrome 45.0.2454.101 (linux).
there's not data now, but it's still better than before

pat-son:
Works fine for me
![screen shot 2015-11-13 at 2 00 59 pm](https://cloud.githubusercontent.com/assets/8462638/11155018/d046ff08-8a10-11e5-9d6d-a7850d84892a.png)

tylerpuleo:
Works fine for me as well

PR #2302, opened 2015-11-12
Fixed issue ps2276 - project 891 breaks vis rendering

pat-son:
#2276 
Basically, one of the objects in the scatterplot vis was undefined, so when the script tried to access it, it threw an Uncaught TypeError, so it halted the rendering of the page.


asalani93:
didn't personally test because i didn't have a broken project to test on
also, in the future, put the issue number of the bug that it fixes, add a "needs review" label and assign to someone.  last two are incredibly unimportant and the first one is only a convenience thing. nbd
everything else looks good

PR #2313, opened 2015-11-19
Fixes a test that broke from material design

bdonald25:
No associated bug.
Last release had a test that clicked a check box. Material design was merged at the same time and it broke the way the test found the check boxes. This fixes the test so it finds the check boxes correctly.

### DIFF
--- a/test/integration/number_of_data_sets_test.rb
+++ b/test/integration/number_of_data_sets_test.rb
@@ -1,15 +1,15 @@
 require 'test_helper'
+require 'capybara-screenshot'
 require_relative 'base_integration_test'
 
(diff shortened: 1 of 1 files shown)


asalani93:
Consistently fails
`Failure:
UploadDataTest#test_correct_number_of_data_sets [/home/asalani/Documents/ecg/rSENSE/test/integration/number_of_data_sets_test.rb:13]
Minitest::Assertion: Not on project page.`
Passes when only that test is run.  Fails otherwise.

bdonald25:
@asalani93 You are right, it now passes when only that test is run but still fails when all the tests are run together. That is super annoying.

Issue #2271, opened 2015-10-16
Uploading data set while templating uploads empty rows


asalani93:
**General description:** Choosing to upload a data set when you template fields uploads a blank data set.

**live/dev/localhost:** live
**iSENSE Version:** 7.3
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Ubuntu 15.04 x64
**Browser/Version:** Chrome 45 x64

**Steps to Reproduce:**
* Create a project
* Template the fields from a CSV with data
* Choose to upload the data from the CSV as well
* Observe that there are no data in the new data set

**Associated Image(s):**

bdonald25:
Testers: Template fields by uploading a csv. Make sure the option to create a data set at the same time is checked. After saving the data set should be created and contain data.

bdonald25:
PASSED testing on Chrome 48.0.2564.116 (linux).

pat-son:
PASSED testing on Chrome 48.0.2564.116 (OSX 10.10.5).

PR #2314, opened 2015-11-20
Creating a dataset while templating fields now works correctly.

pat-son:
This fixes [Issue 2271](https://github.com/isenseDev/rSENSE/issues/2271). Before, templating fields then checking "create dataset" would create a dataset with the correct number of rows, but they would be empty. This was due to an error in finishTemplateUpload, in which the newly templated fields were saved to the backend, but this change was not reflected in the project variable already created in the function. So when "swap_with_field_names" was called and tried to iterate through the project object's fields, it would find them empty. 

To fix this, I simply added `@project = Project.find(params[:id])` which retrieves the newly saved fields from the backend. There is a way to fix the issue without calling Project.find again, but that requires rewriting swap_with_field_names.


bdonald25:
Periodically fails test_contributions.

pat-son:
Closing this pull request to retry it in a different one.

Issue #2285, opened 2015-10-31
Error bars are broken


bdonald25:
**General description:** Error bars are broken

**live/dev/localhost:** live
**iSENSE Version:** v7.4.1
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** X11
**Browser/Version:** Chrome 46.0.2490.80

**Steps to Reproduce:** 

393

asalani93:
If someone else decides to do this before I do, the problem is that a function signature was changed without testing error bars, and error bars still uses the old signature.

asalani93:
PASSED testing on Chrome 45.0.2454.101 (Ubuntu 15.04).

PR #2315, opened 2015-11-22
Fix mean with error

bdonald25:
#2285 The function signature changed when data filtering was introduced.


asalani93:
Looks good.  Not sure if you want to merge because every PR is failing that one test.

PR #2325, opened 2015-12-03
Devise2

jdalphon:
Our user model is now handled by devise!. See (https://github.com/plataformatec/devise)

This is not to be pulled before the next release. After the release it should end up on dev to set up mailers and make sure that is all working.


bdonald25:
@jdalphon 
Devise is now on dev for testing. I found a couple issues.

1. Creating a user redirects to rsense-dev.cs.uml.edu/users which gives the beaker man. User creates successfully though.
2. Create a user page has a link to log in way off to the bottom left of the page.
3. Privacy policy link is in the wrong spot.
4. I personally do not like the rectangle around the 'remember me' check-box 
5. The error for creating a user and entering a password that is too short is not the bootstrap themed error that we use everywhere else. It is also not centered.
6. Change email or password link on users page gives beaker man.

jdalphon:
@bdonald25 
Yay, This is why we have a dev server. all of those errors do not occur on my local copy. 

The styling is another issue entirely.

bdonald25:
Have two tabs open, signed in, on project index page.
Sign out on one tab.
Create a project with the other tab.
Dev fails silently and redirects to homepage.
Live redirects to sign in page.

asalani93:
* Most alerts (the ones in the red box below the nav bar) are shown twice.  Example: register with an existing email address.  The error "Email has already been taken" shows twice.  Same with changing your email to something that's invalid.
* Certain invalid emails get different validation errors.  "a" and "a@" get little boxes telling you how to get a valid email. "a@a" tries to make an account, fails, and sends back the error "Enter a valid email address" twice.
* Password length is not mentioned when you go to change a password (until you make one too short, then it tells)
* On the password change page, sometimes the "Password" and "Password confirmation" text boxes have their contents shifted all the way to the right with no padding
* Exploding beaker man when you try to make a user, get a validation error, and then try again.
* Deleting my account says "You need to sign in or sign up before continuing" when I press it.
* After doing the above and going back a page, clicking logout gives the same error.  Middle-clicking doesn't, however.
* Logging in to a deleted account gives no error message, and just silently sends you back to the login page

jdalphon:
@asalani93 I have added an after sign out to redirect to the home page. We need to come up with a specific list of where redirects should go.

The rest should be updated. We also need to decide if we are going to move away from Travis, these failed checks are very annoying.

bdonald25:
#2375 replaces this

Issue #2231, opened 2015-07-29
API returns internal server error for get requests to some projects.


bdonald25:
**General description:** API returns internal server error for get requests to some projects. (Projects that were deleted?)

**live/dev/localhost:** live
**iSENSE Version:** v7.2.2
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** X11
**Browser/Version:** Chrome 44.0.2403.107

**Steps to Reproduce:** http://rsense-dev.cs.uml.edu/api/v1/projects/500

![screenshot from 2015-07-29 12 57 05](https://cloud.githubusercontent.com/assets/4121924/8964070/7912500c-35f1-11e5-97bf-6df5fe9dc062.png)

393

bdonald25:
Testers: This fixes some projects such as isenseproject.org/projects/500 that gave the beaker man because the user who created the project was deleted.

If this project works on dev then the issue was fixed. rsense-dev.cs.uml.edu/projects/500

pat-son:
PASSED testing on Firefox 44.0.2 (Mac OSX Yosemite 10.10.5).

bdonald25:
PASSED testing on Chrome 48.0.2564.109 (linux).

bdonald25:
PASSED testing on Chrome 48.0.2564.109 (linux).

JasonD94:
PASSED testing on Chrome 48.0.2564.97 (Linux Ubuntu 14.04 LTS).

PR #2331, opened 2015-12-28
Fix broken projects

bdonald25:
#2231 Some projects are broken because the User that created the project has been deleted but their projects are still on iSENSE. This is probably a symptom of us hiding projects instead of deleting them.

Example
http://isenseproject.org/projects/500
http://isenseproject.org/api/v1/projects/500

The project page will not display the user that created the project instead of crashing.
The api call will give null as the user instead of breaking.


tylerpuleo:
Code looks good and it passes all the auto tests, but it didn't work when I tested it locally.

Here's my steps:

1. Make a new local branch
2. Fetch and merge Bobby's api500 branch
3. start server
4. Create a new user and a new project by this user
5. Use admin account to delete user
6. Go back to project/theID#
7. I didn't get beaker man I got this:

selection_003

bdonald25:
That is correct behavior. That is telling you it cannot find the project which is correct because it was deleted when the user was. You get that screen because when you run isense locally it runs in debug mode. On live or dev you would get a 404.

There are some projects that exist, but the user that created them do not because we used to hide projects instead of deleting them. This fixes it so those hidden projects can still be viewed rather than a 500 error. If your db is a clone of live go to localhost:3000/projects/500.

tylerpuleo:
I miss understood what the bug was. Good to go.

bdonald25:
Testers: This fixes some projects such as isenseproject.org/projects/500 that gave the beaker man because the user who created the project was deleted.

If this project works on dev then the issue was fixed. rsense-dev.cs.uml.edu/projects/500

PR #2333, opened 2015-12-30
Manual Entry Pre-Populate DS Name

bdonald25:
#2338 Pre-populates dataset name with username/contrib name + date and time in the format (Bobby 12/30/2015 4:29pm)


tylerpuleo:
Good to merge

PR #2336, opened 2016-01-20
Group by Number Fields for Pie Chart

bdonald25:
Project has 3 number fields -> Wins, losses, and Ties
This pull request will allow you to make a pie chart to compare these 3 fields. 

It works by adding a new group by option just for Pie chart. While grouping by number fields, each slice of the pie represents a seperate number field instead of a portion of a single number field.


bdonald25:
#2337 Replaces this pull request

Issue #1852, opened 2014-12-02
Group by numerical fields


jaypoulz:
**General description:**
A teacher thought it would be useful for bar charts if you could group by the y-fields themselves. This sometimes makes sense for numerical fields, and should be considered in a grouping overhaul.

**live/dev/localhost:** N/a
**iSENSE Version:** N/a
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** N/a
**Browser/Version:** N/a

**Steps to Reproduce:** N/a

**Associated Image(s):** N/a

bdonald25:
Testers: Change the group by options to group by Contributor. Make sure it works for new and old vises, new and old default vises, and new and old saved vises.

Also make sure that all the old saved vises and default vises are the same on live and dev. They should have all the same options for all their tools other than the option to group by contributor. If different things are checked off for a saved vis on live and dev then that is a bug.

pat-son:
PASSED testing on Firefox 44.0.2 (Mac OSX Yosemite 10.10.5).

bdonald25:
PASSED testing on Chrome 48.0.2564.109 (linux).
Should not be an option for photo vis though.

bdonald25:
PASSED testing on Chrome 48.0.2564.109 (linux).
Should not be an option for photo vis though.

JasonD94:
PASSED testing on Chrome 48.0.2564.97 (Linux Ubuntu 14.04 LTS).

bdonald25:
This pull request needs to be looked at a little more so I am going to hold off until next release.

JasonD94:
PASSED testing on Chrome 48.0.2564.109 (linux).

PR #2337, opened 2016-02-02
Groups by Numbers and Contributors

bdonald25:
#1852  Group by Numbers (For Pie Chart Only)
#2088 Group by Contributors

### DIFF
--- a/app/assets/javascripts/visualizations/highvis/bar.coffee
+++ b/app/assets/javascripts/visualizations/highvis/bar.coffee
@@ -142,8 +142,10 @@ $ ->
 
       drawControls: ->
         super()
-
-        @drawGroupControls(data.textFields)
+        # Remove group by number fields, only for pie chart
+        groups = $.extend(true, [], data.textFields)
+        groups.splice(2, 1)
+        @drawGroupControls(groups)
         @drawYAxisControls(globals.configs.fieldSelection,
           data.normalFields.slice(1), false)
         @drawToolControls(true, true)
--- a/app/assets/javascripts/visualizations/highvis/bar.coffee
+++ b/app/assets/javascripts/visualizations/highvis/bar.coffee
@@ -144,7 +144,7 @@ $ ->
         super()
         # Remove group by number fields, only for pie chart
         groups = $.extend(true, [], data.textFields)
-        groups.splice(2, 1)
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/visualizations/highvis/bar.coffee
```diff
@@ -142,8 +142,10 @@ $ ->
 
       drawControls: ->
         super()
-
-        @drawGroupControls(data.textFields)
+        # Remove group by number fields, only for pie chart
```
tylerpuleo: What's the rational behind adding the feature to base vis and then removing it from all other vises as opposed to just adding it to pie.coffee?

app/assets/javascripts/visualizations/highvis/highmodifiers.coffee
```diff
@@ -278,15 +300,21 @@ $ ->
...
+          if data.fields[f].fieldID != -1
+            groups.push(data.fields[f].fieldName)
+      else
+        for dp in @dataPoints
+          if dp[gIndex] isnt null
+            result[String(dp[gIndex]).toLowerCase()] = true
 
-      groups = for keys of result
-        keys
+        groups = for keys of result
```
tylerpuleo: looks like an indentation error

app/assets/javascripts/visualizations/highvis/highmodifiers.coffee
```diff
@@ -278,15 +300,21 @@ $ ->
...
+          keys
```
tylerpuleo: same here

tylerpuleo:
Read my comments. Looks good otherwise. Migration was successful, none of the projects I used to test were broken on dev new or old.

bdonald25:
Those indentations are intentional. Those lines are meant to be in the else part of the if statement. Those lines are getting the groups based on the data of the field that is being grouped by.  For example, the different names from the data in the data set name field. Group by numbers doesn't work like that, it gets its groups from the field names instead of the data of the field that is being grouped. 

Also, I am not adding it to base vis and then removing it from the other visualizations. The group by fields are treated like any other field. They are text fields. Before, all the visualizations would call @drawGroupControls(data.textFields). Instead of adding all the text fields to the group by menu, I am displaying all the text fields except group by numbers for visualizations other than pie.

tylerpuleo:
Good to merge then

PR #2339, opened 2016-02-04
Capitalization lost in "Group by" function

### LINKED ISSUE #2265: Capitalization lost in "Group by" function
When text is entered into a data set, the capitalization pattern is lost in the "Group by" function.  This is important for classroom use, as it often results in proper nouns without capitalization.

For example see:
http://isenseproject.org/projects/1386/data_sets/12477,12468,12467
and group by country.

JasonD94:
Should fix #2265 

I created a project on Dev for testing: http://rsense-dev.cs.uml.edu/projects/1165


bdonald25:
@JasonD94 This looks good but it now has a merge conflict after I merged the other pull requests.

PR #2346, opened 2016-02-19
Creating a dataset while templating fields now works correctly (Take 2).

pat-son:
This fixes Issue #2271. Before, templating fields then checking "create dataset" would create a dataset with the correct number of rows, but they would be empty. This was due to an error in finishTemplateUpload, in which the newly templated fields were saved to the backend, but this change was not reflected in the project variable already created in the function. So when "swap_with_field_names" was called and tried to iterate through the project object's fields, it would find them empty.

To fix this, I simply added @project = Project.find(params[:id]) which retrieves the newly saved fields from the backend. There is a way to fix the issue without calling Project.find again, but that requires rewriting swap_with_field_names.


test/integration/template_fields_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    find(:css, '#template_file_form').attach_file('file', csv_path)
+    assert page.has_content?('Please select types for each field below.'), 'Didn\'t find upload button'
+    click_on 'Submit'
+    visit project_path(@project)
+    find('.data_set_edit').click
+    # this test checks the dataset edit page instead of the vis because
+    # the vis was not rendering properly in capybara.
+    assert page.has_content?('-71.36945'), 'Data wasn\'t uploaded.'
+  end
+end
```
bdonald25: you need a new line at the end

test/fixtures/projects.yml
```diff
@@ -116,3 +116,7 @@ map_vis_disable:
 lots_of_data_sets:
   title: Lots of Data Sets
   user: nixon
+
+template_fields_project:
+  title: Test for templating fields
+  user: nixon
```
bdonald25: you need a new line at the end

pat-son:
Oops, my new install of Sublime does not append new lines. Fixing it now.

pat-son:
Closed because [another request](https://github.com/isenseDev/rSENSE/pull/2357) includes these fixes.

Issue #2335, opened 2016-01-20
Can't create and delete fields at the same time


asalani93:
**General description:**
When you modify a project's fields on the site, you cannot create and delete fields from a project at the same time.  Any fields created at the same time as you delete fields will be ignored.

**live/dev/localhost:** live/dev/localhost
**iSENSE Version:** does anyone even use this
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** computer
**Browser/Version:** internet

**Steps to Reproduce:**
* create a project
* add some fields (like A, B, C or something)
* save the fields
* go back to the field edit page, delete the field named C and a field named D
* save the fields again
* observe that C is gone but D wasn't created
* apparently only works with numbers and timestamps?

**Associated Image(s):**

bdonald25:
@asalani93 It works fine for me. Have you pulled recently?

asalani93:
Removed "cannot reproduce" because we reproduced it last week.

pat-son:
The reason this happens is the edit fields page doesn't discriminate between how many fields you've added and how many you've deleted for each type; it just has "numberCount", textCount", etc. and increments that number when adding a field and decrements it when deleting. So when you delete and add a field at the same time, the number stays at 0 and the for-loop that adds fields doesn't run.

pat-son:
It actually shouldn't have worked for numbers either, but all this time the variable getting decremented was numberCount, but that was the only place that it is used. Everywhere else in the code uses "numCount", and that's what is passed to the projects_controller when saving the fields.

bdonald25:
Testers: Create a new project -> Add one of each field and prepend the names with "old" (ie, old_number, old_text, old_timestamp ect.) -> save -> edit fields -> do the same thing as before but prepent the word new (new_number, new_text ....) -> delete the old fields at the same time before saving -> save

If all the old fields are gone AND all the new fields were created successfully then this passes.

tylerpuleo:
FAILED: Isn't fixed for timestamps. To reproduce add a timestamp, save, click edit, delete then re-add timestamp, then save timestamp isn't there.

bdonald25:
PASSED testing on Chrome 48.0.2564.116 (linux).

pat-son:
PASSED testing on Chrome 48.0.2564.116 (OSX 10.10.5).

PR #2351, opened 2016-02-24
Simultaneously creating and deleting fields of the same type now works correctly

pat-son:
This is a fix for issue #2335. I fixed it by deleting some decrement operations that were causing the projects_controller file to not recognized newly added fields when other fields of the same type were deleted. Removing the decrements does not seem to have any negative repercussions, as deleted fields will get deleted regardless of the numCount and textCount variables, but if you catch anything, let me know.


tylerpuleo:

test/fixtures/projects.yml
```diff
@@ -116,3 +116,7 @@ map_vis_disable:
 lots_of_data_sets:
   title: Lots of Data Sets
   user: nixon
+
+fields_test:
+  title: Field Test for Simultaneous Add and Delete
+  user: nixon
```
bdonald25: @pat-son Another Nixon project here on this pull request that is causing test_contributions to fail.

PR #2352, opened 2016-02-25
Gets Circle CI up and running.

bdonald25:
This pull request gets Circle ci up and running. It does not fix all our problems but the tests that fail appear to actually be ones that are in need of fixing.


tylerpuleo:
@bdonald25 Good to merge

Issue #2349, opened 2016-02-23
Items in Table search drop-down menu get highlighted twice


pat-son:
General description: When highlighting a menu item, we don't need to add the "ui-state-hover" class to the text, because the box itself is already highlighted. See image.

live/dev/localhost: live
iSENSE Version: v7.7.0
Logged in (Y or N): Y
Admin (Y or N): N

OS: Ubuntu
Browser/Version: Chrome

Steps to Reproduce: Hover over a dropdown item on the table vis, and watch as it is highlighted twice.

![dropdown_bug](https://cloud.githubusercontent.com/assets/8462638/13265344/6f962078-da40-11e5-9071-3761f4e06d55.png)

bdonald25:
Testers: Make sure the items in the drop-down menu shown in the image above no longer double highlight. The drop-down menu is located on the table visualization.

tylerpuleo:
PASSED

bdonald25:
PASSED testing on Chrome 48.0.2564.116 (linux).

pat-son:
PASSED testing on Chrome 48.0.2564.116 (OSX 10.10.5).

PR #2354, opened 2016-02-26
Fixes double highlighting bug (#2349)

VictoriaAlbanese:
#2349


tylerpuleo:

bdonald25:
@VictoriaAlbanese Looks good. Nice first bug fix!

Issue #2291, opened 2015-11-09
Canceling printing for a vis breaks the vis.


bdonald25:
**General description:** Canceling printing for a vis breaks the vis.

**live/dev/localhost:** live
**iSENSE Version:** v7.4.1
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** X11
**Browser/Version:** Chrome 46.0.2490.80

**Steps to Reproduce:** Click Print in the vis tools for any vis. Cancel printing. Watch the vis break.

393

bdonald25:
Testers: Go to a visualization -> try to print it from the tools on the left -> cancel the print -> verify the visualization looks sane and the page is not broken

bdonald25:
PASSED testing on Chrome 48.0.2564.116 (linux).

pat-son:
PASSED testing on Chrome 48.0.2564.116 (OSX 10.10.5).

PR #2355, opened 2016-02-26
Canceling print no longer breaks vis

pat-son:
This fixes issue #2291. The root of the problem is an [error in highcharts](https://github.com/highcharts/highcharts/issues/1093) that has gone unfixed for years. The workaround I made was to manually resize the page after the print dialog closes.


tylerpuleo:

Issue #2358, opened 2016-03-02
The row that is selected in the table is a terrible color.


bdonald25:
**General description:** The row that is selected in the table is a terrible color. I am not sure what color it should be. 

You can reproduce this by clicking on a cell in the table visualization.

![table](https://cloud.githubusercontent.com/assets/4121924/13446529/55a910c4-dfe2-11e5-83ed-bd7106d0c7ee.png)

**live/dev/localhost:** live
**iSENSE Version:** v7.7.0
**Logged in (Y or N):** N/A
**Admin (Y or N):** N/A

**OS:** X11
**Browser/Version:** Chrome 48.0.2564.116

**Steps to Reproduce:** Table Vis.

393

bdonald25:
PASSED testing on Chrome 49.0.2623.87 (linux).

tylerpuleo:
PASSED Chrome (mint)

pat-son:
PASSED testing on Chrome 49.0.2623.87 (OSX 10.11.3).
I'm not really sure about the blue border, but it's better than before.

PR #2362, opened 2016-03-04
fixed bug #2358 bad table select highlight color

VictoriaAlbanese:
#2358
The bug here was that when you selected a row of the table, a border was created around the row in a really ugly light yellow color.  The border was around all of the row/column lines in that row except for the top line.  I overrode the bad table selection highlighting by overriding the problem class with a class that adds an even, dark blue border just around the row itself.

Since there weren't any specific requests for what the correct table select should look like, and I couldn't find any similar outline selection anywhere else on the website to match it with, I just kind of used my judgement as to what looked good.  It's definitely better than the yellow, but if you need me to change the outline color, thickness, background-color, etc just let me know and I can update that too.

### DIFF
--- a/app/assets/stylesheets/visualizations.css.scss
+++ b/app/assets/stylesheets/visualizations.css.scss
@@ -538,6 +538,10 @@ ul#sopt_menu {
   }
 }
 
+tr.ui-widget-content.jqgrow.ui-row-ltr.ui-state-highlight {
+  border: 2px solid #263960;
+}
+
 body[data-page-name='visualizations/displayVis'] .visualizations-controller,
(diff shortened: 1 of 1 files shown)


bdonald25:
screenshot from 2016-03-08 14-24-51
Thoughts?

pat-son:
Looks okay to me, but I'm not really a design guy.

doug-salvati:
The border is good but the background-color looks to be the same yellow color. Victoria, didn't you have it gone at one point?

bdonald25:
Looks better than before. I am going to merge this for release tomorrow. If we want to make changes then we can after release.

Issue #2179, opened 2015-06-09
Add a period option to timeline


asalani93:
**General description:** When comparing periodic data on the timeline vis (like temperatures or something), it would be useful to be able to visual this data periodically, ie: it displays data starting from January 1st to December 31st, and then once the next year starts it wraps around back to January 1st and starts on the left side of the vis again.
There could be options for different periods (yearly, monthly, daily, etc...), too.
One use case for this is the boat house: the tablet applications wouldn't need to set the year to 2010 just to have the data be comparable on timeline.

**live/dev/localhost:**
**iSENSE Version:**
**Logged in (Y or N):**
**Admin (Y or N):**

**OS:**
**Browser/Version:**

**Steps to Reproduce:**

**Associated Image(s):**

jdalphon:
+1

bdonald25:
Testers: Make sure time period options work and grouping by time period.

VictoriaAlbanese:
Couldn't find any bugs with the time period stuff
-Passed

VictoriaAlbanese:
PASSED testing on Chrome 48.0.2564.116 (linux).

doug-salvati:
Period appears for all visualizations but only seems to do anything on timeline.

I assume when I pick "monthly" it is showing all months on top of each other.  It would be nice to be able to filter by month.
As far as I can tell, it is not possible to save a Date filter without getting something like "Date contains Invalid Date"

pat-son:
PASSED testing on Chrome 50.0.2661.86 (OSX 10.11.3).
Wow, this is a great feature. Whoever wrote this is pretty good.

pat-son:
@doug-salvati The period option for other vises is for the group-by option.

PR #2380, opened 2016-04-28
View timeline data in a periodic range.

pat-son:
#2179 Brand new feature: Users can view timeline data by period. Options include yearly, monthly, weekly, daily, and hourly. When the period option is set, the level chosen is abstracted away from the data. e.g. When choosing the Yearly option, the dates October 14th 1994 and October 14th 2016 will have the same x-value on the chart.

Integration tests still need to be written for this feature, which will be done later.


bdonald25:
Testers: Make sure time period options work and grouping by time period.

PR #2390, opened 2016-05-16
Create project fab button

bdonald25:
@pat-son This still needs tests but here is my code so it is not lost forever.


pat-son:
Closing to work on locally.

Issue #2397, opened 2016-05-17
return email addr for reset link should be changed


fgmart:
presently it's `please-change-me-at-config-initializers-devise@example.com`

let's change it to `isenseproject@gmail.com`.

pat-son:
Testers: Try resetting your password and see what the return address is in the email.

pat-son:
PASSED testing on Chrome 50.0.2661.102 (Ubuntu 16.04).

PR #2398, opened 2016-05-17
change email for forgot password

bdonald25:
#2397


pat-son:
For some reason this fails bug_report_test.rb

Issue #2413, opened 2016-05-31
Code Injection on Vis Page


pat-son:
General description: By entering valid HTML into a text field/field name, a user can get arbitrary code to run when a user views the table/summary vis. This is because we do not sanitize data and field inputs.

live/dev/localhost: All 3
iSENSE Version: 8.1
Logged in (Y or N): Either
Admin (Y or N): Either

OS: Any
Browser/Version: Any

Steps to Reproduce: Open a new project, add some fields and data. Change a field name so that it has valid JavaScript between two <script> tags, then view the table vis. JavaScript executes.

pat-son:
Testers: Make sure that you cannot inject code into any part of the website. That includes field names, text data, and even text formula data.

pat-son:
PASSED testing on Chrome 50.0.2661.102 (Ubuntu 16.04).
Code injection is no longer allowed, but trying to inject code into the field name causes the manual entry page to break in some cases. Will look into it more later.

PR #2414, opened 2016-06-01
Users can no longer inject client-side code into the website.

pat-son:
For #2413.
Input sanitization is done at the database level, so even data coming from APIs and CSVs gets sanitized.
All HTML tags are stripped away, which is fine because I can't think of a use case when someone would need \<brackets\> in their dataset.


bdonald25:
RIP my favorite bug

pat-son:
Sorry to ruin your fun, @bdonald25, but it had to be done.

Issue #2433, opened 2016-06-23
"Mine" option in the Data set checkbox dropdown is not working


pat-son:
**General description:** If you go to the data sets panel on a project page for which you own data sets, the dropdown option for checking all of your datasets ("Mine") does not work anymore. The specific error is "Uncaught Type error: cannot read property "check" of undefined", pointing to [line 125 in show.js.coffee](https://github.com/isenseDev/rSENSE/blob/0cf70c82245e440f66a6bdc5fd285ac6f39b8c7a/app/assets/javascripts/projects/show.js.coffee#L125).

**live/dev/localhost:** All 3
**iSENSE Version:** v8.1.1
**Logged in (Y or N):** Yes
**Admin (Y or N):** No

**OS:** Ubuntu 16.04
**Browser/Version:** Chrome 50.0.2661.102

**Steps to Reproduce:** Create a project with a data set or two, or just view an existing project. In the "Data Sets" panel on the project page, click the dropdown in the top right and select "Mine". Notice how nothing gets selected. Also note the error in the Javascript console.

pat-son:
Testers: In a project with mixed owned/unowned data sets, use the "mine" selection and make sure it works as intended. Make sure it still allows you to visualize, export, and delete.

pat-son:
PASSED testing on Chrome 50.0.2661.102 (Ubuntu 16.04).

PR #2448, opened 2016-07-05
Mine button fix on data sets page

mck529:
Issue #2433 
Fixing issue with mine button 
Changed where the ".mine" extension was added (which is how the button found which checkboxes to click) - was previously being attached to an object which was undefined

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -181,8 +181,8 @@
                     </td>
                     <td>
                       <div class="ds_selector" >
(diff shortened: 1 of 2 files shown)


app/views/projects/show.html.erb
```diff
@@ -181,8 +181,8 @@
                     </td>
                     <td>
                       <div class="ds_selector" >
-                        <label id="lbl_<%=s.id%>" class="mdl-checkbox mdl-js-checkbox mdl-js-ripple-effect pull-right" for="ds_<%=s.id%>">
-                          <input type="checkbox" id="ds_<%=s.id%>" class="mdl-checkbox__input <%if s.user_id == current_user.try(:id)%> mine<%end%>" checked>
```
pat-son: Removing this line breaks the "delete selected" button - needs to be brought back

Issue #2489, opened 2016-09-09
Typo in visualizations.css.scss : "whitespace" invalid property


mck529:
**General description:**
Found in line 377 of visualizations.css.scss
"whitespace" is not a valid property. Should be spelled "white-space"
Due to the property being invalid, this line does nothing. 
Line should be either fixed or removed.

**Associated Image(s):**
![whitespacetypo](https://cloud.githubusercontent.com/assets/19916403/18398091/23338c8e-7699-11e6-81ef-3d187fd063fe.png)

pat-son:
Testers: Add a filter on a project and make sure it looks okay.

doug-salvati:
PASSED testing on Chrome 53.0.2785.143 (OSX 10.11.6).

mck529:
PASSED testing on Chrome 53.0.2785.101 (linux).
Filters still display properly

VictoriaAlbanese:
PASSED testing on Chrome 48.0.2564.116 (linux).

PR #2496, opened 2016-09-14
Removed line with typo (visualizations.css.scss 407)

mck529:
**For Issue #2489** 
Line was not functional to begin with.
Simply removed line since it was not operational


tylerpuleo:
@pat-son This probably doesn't need to be tested. Lets just merge.

PR #2501, opened 2016-09-20
Commented out the html5 validation tests

pat-son:
Will uncomment them once the validator gets fixed on their end.


tylerpuleo:
Good to go. I'm merging it.

Issue #2547, opened 2016-11-10
Teacher request: More error bar information


pat-son:
What do the error bars mean? We should put that information in the error bar tooltips.

doug-salvati:
Testers: make sure the standard deviation in the error bars looks like it matches the one on the Summary page. Also, make sure "Mean with Deviation" shows up on the tooltip

pat-son:
PASSED testing on Chrome 54.0.2840.100 (linux).

VictoriaAlbanese:
PASSED testing on Chrome 54.0.2840.71 (linux).
Everything looks fine! 
It would be cool to include the standard deviation itself on the tooltip, though.

doug-salvati:
PASSED testing on Safari 10.0.3 (mac).

PR #2556, opened 2016-11-23
Corrects StdDev formula and tooltip for error bars

doug-salvati:
For issue #2547 

- Error bars simply indicate the positions of plus and minus one standard deviation.  Since standard deviation is not the same as error, the option is now called "Mean with Deviation" and the tooltip reflects appropriately.
- There was a small error in the formula for standard deviation in error bars.  It is fixed to be consistent with the Summary tab.


doug-salvati:
Testers: make sure the standard deviation in the error bars looks like it matches the one on the Summary page.  Also, make sure "Mean with Deviation" shows up on the tooltip

Issue #2542, opened 2016-11-05
Filtering out data from Bar vis results in blank bars being drawn


pat-son:
**General description:** When you filter out data points, the bars will go away but the labels persist and there's still space allocated on the graph for them.

**live/dev/localhost:** live
**iSENSE Version:** 8.2.3
**Logged in (Y or N):** Y
**Admin (Y or N):** N

**OS:** Any
**Browser/Version:** Any

**Steps to Reproduce:**
https://drive.google.com/file/d/0ByhfLrDH6PhIYTJLTk1iVll4YnM/view

doug-salvati:
Testers: Make sure the behavior shown in the video no longer happens.

pat-son:
PASSED testing on Chrome 54.0.2840.100 (linux).

VictoriaAlbanese:
PASSED testing on Chrome 54.0.2840.71 (linux).
Works perfectly

doug-salvati:
PASSED testing on Safari 10.0.3 (mac).

PR #2557, opened 2016-11-28
Only displays groups in legend or chart if they have any data.

doug-salvati:
For issue #2542


doug-salvati:
Testers: make sure this behavior does not happen anymore: https://drive.google.com/file/d/0ByhfLrDH6PhIYTJLTk1iVll4YnM/view

Issue #2555, opened 2016-11-18
Values become undefined when histogram blocks merge


doug-salvati:
**General description:** When bin size increases, each data point no longer occupies its own block.  The data point will report 'undefined' for the value because it corresponds to many points.  Perhaps it should display the average on the tooltip (full list on detailed tooltip?), or nothing at all.
**live/dev/localhost:** live
**iSENSE Version:** 8.2.4
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** Any
**Browser/Version:** Any

**Steps to Reproduce:**  [https://isenseproject.org/projects/34/data_sets/](https://isenseproject.org/projects/34/data_sets/) for example: change bin size to 5 and observe that the serving size is indicated on the tooltip.  Then, change bin size to 100 and observe that serving size is undefined.

![image](https://cloud.githubusercontent.com/assets/17438678/20445475/335ab06a-ada3-11e6-84aa-7a091f36d6f9.png)
![image](https://cloud.githubusercontent.com/assets/17438678/20445446/18114350-ada3-11e6-888b-000d67e147e3.png)

doug-salvati:
Testers: Make bins small and check that the values are shown on tooltip. Then make them large enough to consolidate and make sure that tooltips display the # of items in the bin, instead of an undefined value.

The Fast Food project has good amount of data for testing this.

pat-son:
PASSED testing on Chrome 54.0.2840.100 (linux).

VictoriaAlbanese:
PASSED testing on Chrome 54.0.2840.71 (linux).
Works perfectly

doug-salvati:
PASSED testing on Safari 10.0.3 (mac).

Issue #1470, opened 2014-04-01
Enhance axis labeling/mouseover tooltips for histogram


ivanrudnicki:
Users would have an easier time understanding histograms, I believe, if the x axis labeling were clearer or the tooltips more detailed. As it is now, depending on the bin size, all or just a handful of the bars in a histogram have labels on the x axis. Even for a bar with a label, the user has to mentally calculate the lower and upper values represented in the stack: lower value =  label - (bin size/2) and upper value = label + (bin size/2). This is kind of hard. I expect that enhancing the high charts axis labels might be difficult, but could the tooltips provide more detail on the range of values represented in each stack?

pat-son:
We will either make the x-axis ticks positioned more appropriately or add a new label on the tooltip. Definitely gonna do this soon.

VictoriaAlbanese:
Would something like this work?

Fieldname: ...
Group ID: ...
Bin #: ...
Bin Range: (lower bound) - (upper bound)
Data Points in Bin:

I think adding a range option would solve your problem, and that rephrasing "# occurances" to "data points in bin" might be a bit more descriptive to younger audiences who aren't familiar with histograms in the first place.  

Also, currently the last field in the tooltip looks like this: 
(group id): 1
which doesn't seem to make much sense.  It might make more sense to include it on the tool tip as I did above, instead of saying (I presume) that this represents 1 data point taken from whatever group.

doug-salvati:
In the process of working on #2555, I was annoyed by this too began reworking the format like this.  I'm redoing the tooltips to show "Data Points in Bin" when there are multiple data points consolidated into a block, and shows the actual value if there is only one data point in the bin.

I would like to add the range feature you implemented.  We should combine our code for the design to close this, and I'll PR the logic behind #2555 later on down the road...

![imageedit_2_4405756149](https://cloud.githubusercontent.com/assets/17438678/22299457/24085926-e2f2-11e6-8b45-aad823c36611.jpg)

doug-salvati:
Testers: make sure range is correct in the tooltip, and that count per bin displays as well (only happens if the bin is big enough)

pat-son:
PASSED testing on Chrome 54.0.2840.100 (linux).

VictoriaAlbanese:
PASSED testing on Chrome 54.0.2840.71 (linux).
Works perfectly

doug-salvati:
PASSED testing on Safari 10.0.3 (mac).

PR #2571, opened 2017-01-26
Clearer tooltips with no "undefined"s

doug-salvati:
For #1470 and #2555
Clearer tooltips for histograms, including range feature and displays count instead of value in the case where histogram blocks are consolidated because bins are large compared to amount of data points.

Note: whether the bins are consolidated is determined based on whether they have a value defined for them

### DIFF
--- a/app/assets/javascripts/visualizations/highvis/histogram.coffee
+++ b/app/assets/javascripts/visualizations/highvis/histogram.coffee
@@ -63,17 +63,18 @@ $ ->
               
               xField = @series.xAxis.options.title.text
               idx = data.fields.map((x) -> fieldTitle(x)).indexOf(xField)
-              str  = "<div style='width:100%;text-align:center;color:#{@series.color};'> "
-              str += "Bin #{@x}<br>"
+              str  = "<div style='width:100%;text-align:center;'> "
+              str += "<b><u>Bin #{@x}</u></b><br>"
               str += "Contains #{@total} Items<br>"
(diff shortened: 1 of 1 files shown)


doug-salvati:
Testers: Make bins small and check that the values are shown on tooltip.  Then make them large enough to consolidate and make sure that tooltips display the # of items in the bin, instead of an undefined value.

Also, make sure that the tooltip displays the range of the bin correctly.

The Fast Food project has good amount of data for testing this.

app/assets/javascripts/visualizations/highvis/histogram.coffee
```diff
@@ -60,15 +60,26 @@ $ ->
...
-                str += "<tr><td><div style='color:#{@series.color};'> #{@series.name}:</div></td>"
-                str += "<td>#{@y}</td></tr>"
+              str  = "<div style='width:100%;text-align:center;'> "
+              str += "<b><u>Bin #{@x}</u></b><br>"
+              str += "Contains #{@total} Items<br>"
+              str += "Within the Range #{@x - document.getElementById("bin-size").value / 2}"
+              str += "- #{@x + document.getElementById("bin-size").value / 2}</div><br>"
+              str += "<table>"
+              str += "<tr><td style='text-align: right'>Group :&nbsp;</td>"
+              str += "<td style='color:#{@series.color};'>#{@series.name}</td></tr>"
```
pat-son: remove the commented out code and recommit, mostly just so that circle will run on this PR

pat-son: nvm

Issue #2582, opened 2017-01-31
Show standard deviation on bar mean with error tooltip


VictoriaAlbanese:
**General description:** For the bar chart, when using the "Mean with error" setting, which shows the error range for one standard deviation, it would be nice if the tooltip reported the standard deviation.

**live/dev/localhost:** live
**iSENSE Version:** v8.2.5
**Logged in (Y or N):** N
**Admin (Y or N):** N

**OS:** Linux x86_64
**Browser/Version:** Chrome 54.0.2840.71

**Steps to Reproduce:** Mouse over any bar in the bar chart, making sure under tools that the "Mean with Error" analysis type is set (image showing this below)

![error_prefix_tooltip](https://cloud.githubusercontent.com/assets/12720180/22478952/053d2e1e-e7ba-11e6-9c3e-e1e08c59c9b3.png)

3532

VictoriaAlbanese:
Also removing manual calculation of variance and stdDev from the code and using a function predefined in a separate file

VictoriaAlbanese:
Testers: make sure the standard deviation is being displayed in the bar tooltips and that the upper and lower bounds seem correct for the error bar tooltips.

doug-salvati:
FAILED testing on Chrome 54.0.2840.90 (linux).
Looks like this is what I'm getting...
    Upper bound = value + 2 * stddev
    Lower bound = value

VictoriaAlbanese:
PASSED testing on Chrome 54.0.2840.71 (linux).
New fix passed (latest pull request)

doug-salvati:
PASSED testing on Chrome 54.0.2840.90 (linux).

PR #2584, opened 2017-02-02
Show standard deviation on bar mean with error tooltip

VictoriaAlbanese:
Fixed bug #2582 
StdDev was calculated manually instead of with the standard function in a different file, so I fixed that; also, I added the standard deviation to the bar tooltip, and tried to improve the error bar tooltip as well.  Pictures attached below.

new_bar_tooltip
new_error_bar_tooltip

### DIFF
--- a/app/assets/javascripts/visualizations/highvis/bar.coffee
+++ b/app/assets/javascripts/visualizations/highvis/bar.coffee
@@ -88,11 +88,14 @@ $ ->
                   str += "color:#{@point.color};margin-bottom:5px'> "
                 str += "<b><u>Group : #{@point.name}</u></b></div>"
                 str += "<table>"
-                str += "<tr><td style='text-align: right'>Analysis Type :&nbsp</td><td>#{self.analysisTypeNames[self.configs.analysisType]}</td></tr>"
-                str += "<tr><td style='text-align: right'>#{@point.field} :&nbsp</td><td>#{@y}</td></tr>"
-                str += "<tr><td style='text-align: right'>Standard Deviation :&nbsp</td><td>± #{@point.stdDev}</td></tr>"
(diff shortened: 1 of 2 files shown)


VictoriaAlbanese:
Fixing my tests that did not pass (coffeelint).
Also reverting my last commit that calculation of mean was not useless.

app/assets/javascripts/visualizations/highvis/bar.coffee
```diff
@@ -163,9 +183,12 @@ $ ->
         for fid in data.normalFields when fid in fieldSelection
           for gid in sortedGroupIDs when gid in data.groupSelection
             xCoord = (xCluster - 1.5) + pos * interval
+            dp = globals.getData(true, globals.configs.activeFilters)
```
pat-son: set `dp` outside of the for loops because it does not change in each loop

app/assets/javascripts/visualizations/highvis/bar.coffee
```diff
@@ -227,7 +250,7 @@ $ ->
               if barData.length == 1
                 # Error bar must be present or it causes issues with other bars
                 thisError = {
-                  name: "Error for #{name}"
+                  name: "#{name}"
```
pat-son: bars with only one value have NaN in tooltip for error bars. Needs a value for stdDev.

app/assets/javascripts/visualizations/highvis/bar.coffee
VictoriaAlbanese: done

app/assets/javascripts/visualizations/highvis/bar.coffee
VictoriaAlbanese: done

Issue #2578, opened 2017-01-27
Fix API server load issue


tylerpuleo:
Limit API requests is most likely the reason. More info to come.

PR #2587, opened 2017-02-08
Use features of active record to reduce queries

scytacki:
#2578
The features used here are includes, inverse_of, and pluck.
- Includes is used to prevent 1+N queries. In this case each DataSet was making a separate query its user.
- Inverse_of helps ActiveRecord reuse existing objects. So in this case the project object will be the same as data_set.project.
- Pluck is an efficient way to return just the value of one field.

On my local instance of iSENSE it reduced a call to /api/v1/projects/1?recur=true request with 608 datasets from 3500ms to around 550ms. 

Most of the extra queries were hitting the query cache, however even hitting the cache takes extra time especially when it results in creating a new model object.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -62,7 +62,7 @@ class ProjectsController < ApplicationController
 
     @cloned_project = Project.select(:id, :user_id, :title).where(id: @project.cloned_from).first
     @liked_by_cur_user = Like.find_by_user_id_and_project_id(current_user, @project.id)
-    @data_sets = @project.data_sets.select('id', 'title', 'user_id', 'key', 'created_at', 'contributor_name').search(params[:search])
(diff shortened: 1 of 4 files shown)


scytacki:
The last commit improves the loading of the visualization pages when there are a lot of datasets.

There is another improvement that could be made when the list of datasets is specified for the viz.  Currently this makes individual requests for each dataset. And it also checks the datasets are in the project. It fails if any are not in the project. Is that necessary? Could it just filter out those that are not in the project? Here is the code:
```
      dsets.each do |id|
        begin
          dset = DataSet.find_by_id(id)
          if dset.project_id == @project.id
            @datasets.push dset
          else
            fail 'data set does not belong to project'
          end
        rescue
          respond_to do |format|
            format.html { redirect_to '/404.html' }
            format.json { render json: { errors: ['File not found.'] }, status: 404 }
          end
          return
        end
      end
```

I could be be replaced with:
```
@datasets = DataSet.where(:id => dsets, :project_id => @project.id)
```

If necessary the length of @datasets could be checked and the not found error could be thrown if it is 0.

Issue #2594, opened 2017-02-13
DataSets with MediaObjects slow down the visualization


scytacki:
I looked into some continued slowness for the VisualizationsController#displayVis action.  
If there are no MediaObjects and the datasets param is not used then it is very fast.
If several datasets have MediaObjects then it gets much slower.  The issue is caused by the recursive loading the MediaObject information here:
https://github.com/isenseDev/rSENSE/blob/master/app/controllers/visualizations_controller.rb#L310

Below you can see the difference between a request with recursive on and recursive off.  If the extra information about the project and data_set is needed, then I believe this can be improved by using some active record configuration on the models. This way the project and its dependencies won't be loaded again for each media_object.

I will try to work on this, but I wanted to record it here in case I don't.

With 608 datasets and no recursion the request looks like this:
```
app_1     | Started GET "/projects/1/data_sets/" for 172.22.0.5 at 2017-02-13 21:19:43 +0000
app_1     | Processing by VisualizationsController#displayVis as HTML
app_1     |   Parameters: {"id"=>"1"}
app_1     |   Project Load (0.5ms)  SELECT  "projects".* FROM "projects"  WHERE "projects"."id" = 1 LIMIT 1
app_1     |   Field Load (0.5ms)  SELECT "fields".* FROM "fields"  WHERE "fields"."project_id" = $1  ORDER BY field_type ASC, created_at ASC  [["project_id", 1]]
app_1     |   FormulaField Load (0.4ms)  SELECT "formula_fields".* FROM "formula_fields"  WHERE "formula_fields"."project_id" = $1  ORDER BY field_type ASC, created_at ASC  [["project_id", 1]]
app_1     |   DataSet Load (1.4ms)  SELECT "data_sets".* FROM "data_sets"  WHERE "data_sets"."project_id" = 1
app_1     |   User Load (0.5ms)  SELECT "users".* FROM "users"  WHERE "users"."id" IN (1)
app_1     |   MediaObject Load (0.9ms)  SELECT "media_objects".* FROM "media_objects"  WHERE "media_objects"."data_set_id" IN (1, 2, ... <snipped>
app_1     |    (1.0ms)  SELECT COUNT(*) FROM "data_sets"  WHERE "data_sets"."project_id" = 1
app_1     |   User Load (0.4ms)  SELECT  "users".* FROM "users"  WHERE "users"."id" = $1 LIMIT 1  [["id", 1]]
[...]

PR #2595, opened 2017-02-15
Media objects performance

chikeabuah:
#2594

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -137,22 +137,22 @@ class Project < ActiveRecord::Base
       path: UrlGenerator.new.project_path(self),
       hidden: hidden,
       featured: featured,
-      likeCount: likes.count,
+      likeCount: likes.length,
       content: content,
       timeAgoInWords: time_ago_in_words(created_at),
       createdAt: created_at.strftime('%B %d, %Y'),
       ownerName: owner.nil? ? nil : owner.name,
       ownerUrl: owner.nil? ? nil : UrlGenerator.new.user_url(owner),
-      dataSetCount: data_sets.count,
-      dataSetIDs: data_sets.pluck(:id),
-      fieldCount: fields.count,
+      dataSetCount: data_sets.length,
+      dataSetIDs: data_sets.to_a.collect(&:id),
+      fieldCount: fields.length,
       fields: fields.map { |o| o.to_hash false },
-      formulaFieldCount: formula_fields.count,
+      formulaFieldCount: formula_fields.length,
       formulaFields: formula_fields.map { |o| o.to_hash false }
     }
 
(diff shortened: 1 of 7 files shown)


app/models/project.rb
```diff
@@ -137,22 +137,22 @@ def to_hash(recurse = true)
...
       content: content,
       timeAgoInWords: time_ago_in_words(created_at),
       createdAt: created_at.strftime('%B %d, %Y'),
       ownerName: owner.nil? ? nil : owner.name,
       ownerUrl: owner.nil? ? nil : UrlGenerator.new.user_url(owner),
-      dataSetCount: data_sets.count,
-      dataSetIDs: data_sets.pluck(:id),
-      fieldCount: fields.count,
+      dataSetCount: data_sets.length,
+      dataSetIDs: data_sets.to_a.map(&:id),
```
scytacki: This approach will prevent 2 queries. But on a large project, the single query and object loading will probably be a lot slower than the 2 fast queries that don't do any object loading. However, if this is a recursive serialization then this new approach will be faster.   I suspect you made these changes because when the project is loaded via the query in the `visualizations_controller.rb` with all the includes. In that case this approach doesn't result in any more queries but I think it will still result in lots of object loading.  So a lot of memory will be used when there are a lot of media_objects and data_sets.  I'll submit a PR that should be more efficient but I don't have time to test it.

scytacki:
Here is an alternative way to speed things up:
https://github.com/isenseDev/rSENSE/pull/2598

app/models/project.rb
chikeabuah: I disagree :) on a large project with a lot of media sets, eager loading all the information up front would almost definitely be faster than loading piecemeal on each recursive call. Note we're not just talking about 2 fast queries here, altogether without eager loading it is polynomial number of queries (like i mentioned on the issue thread). I'm sacrificing memory here for time, but time is a far more valuable resource than memory. Once everything is loaded into memory, there will be no further object loading downstream.

scytacki: You might be right, it is hard to know with out profiling it.  New Relic is a great way get a sense of this.  But just for sake of argument :).  Let look at some real numbers. This project on isense: https://isenseproject.org/projects/864 has 2,228 datasets and 201 media objects. 

So first the visualization will create 2,228 dataset objects. Then it will create all of those dataset objects again for each media object (when the media object runs to_hash on its project).  So that will be 450,056 dataset objects being created.  If we say each dataset takes 100 bytes (it is probably more than that), that will around 45 MB that isn't too bad, but if it is 1k per object (more likely) it will be 450MB. That is quite of bit of memory for one process to handle and when multiple requests come in at the same time it could be a problem.  If we look at time.

PR #2598, opened 2017-02-17
Should speed up vizes for projects with a lot of media objects

scytacki:
I didn’t have time to test this, but I thought seeing the code would be helpful.


chikeabuah:
will be fixed by removing unnecessary recursion (the root issue)

Issue #2600, opened 2017-02-21
limit spreadsheet uploads to 10 MB


fgmart:
Last week we tried uploading an 8.2 MB spreadsheet and it worked http://isenseproject.org/projects/2797/data_sets/36434

But then when we tried to import 80 MB it failed. After iSENSE thrashed for more than an hour.

We should protect ourselves by limiting upload size and giving an error if user tries to upload a single data set that is too large.

Important: the error message should include the recommendation: "Maximum upload size of a single data set is 10 MB. Please split your file into multiple pieces and upload them individually."

pat-son:
Do you have an example 80 MB file that I can use to test?

fgmart:
https://drive.google.com/open?id=0B41atXlV0KHKTmNNM2VKVURuRW8

tylerpuleo:
TESTERS: Here's a 10MB file to use for testing

https://drive.google.com/open?id=0B4CQhnSpitbrNDZkbGszMjd0M0U

VictoriaAlbanese:
PASSED testing on Chrome 54.0.2840.71 (linux).
Works nicely

pat-son:
PASSED testing on Chrome 54.0.2840.100 (Ubuntu 16.04).

doug-salvati:
PASSED testing on Safari 10.0.3 (mac).
Takes awhile to show the message, but gets the job done.

RistrettoAffogato:
PASSED testing on Safari 10.0.3 (mac).
"Maximum upload size of a single data set is 10 MB. Please split your file into multiple pieces and upload them individually." 
Working as intended.

PR #2608, opened 2017-03-06
Data upload limit

pat-son:
#2600.
Uploads are limited to 10MB.


tylerpuleo:
Looks good. I tried to make a 10MB file to test but it took too long so I gave up. Do you know of any files that are that big that we can test with?

tylerpuleo:
Actually I just made one

Issue #2606, opened 2017-03-01
Set a limit on the number of rows of data that can be uploaded/visualized


pat-son:

pat-son:
Testers: I don't think there are any projects on the dev server with which to test this fix. Oh well

PR #2618, opened 2017-04-05
Added count to DataSet model

pat-son:
For #2606.
Users may now only visualize 100000 data points at a time, and upload a maximum of 100000 data points in each data set. Any more than that and the vis page starts to become unusable with so many data points to parse through, not to mention the strain it puts on the servers.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -173,7 +173,7 @@ MethodCallParentheses:
 # Offence count: 78
 # Configuration parameters: CountComments.
 MethodLength:
-  Max: 130 
+  Max: 150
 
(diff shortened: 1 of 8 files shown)


.rubocop.yml
```diff
@@ -173,7 +173,7 @@ MethodCallParentheses:
 # Offence count: 78
 # Configuration parameters: CountComments.
 MethodLength:
-  Max: 130 
+  Max: 150
```
tylerpuleo: If rubocop yells at you just change it :+1:

pat-son: Yep

PR #2619, opened 2017-04-11
aws environment

chikeabuah:


.ebextensions/deploy.config
```diff
@@ -0,0 +1,118 @@
...
+      sudo chown -R webapp:webapp /home/webapp
+      sudo service nginx reload
+      true
+  "/opt/elasticbeanstalk/hooks/appdeploy/pre/13_ebs_perm.sh":
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
+      #!/usr/bin/env bash
+      if ! mount | grep /dev/xvdh > /dev/null; then
```
mdibello: We're making several assumptions here, and I don't know how valid they are. We're assuming that the device we're looking for is /dev/xvdh and it's mounted on /media/volume1. This might be a reliable assumption to make, but it might not.

.ebextensions/deploy.config
```diff
@@ -0,0 +1,118 @@
...
+        sudo mkfs -t ext3 /dev/sdh
```
mdibello: Is there a reason for using ext3 instead of ext4?

.ebextensions/deploy.config
```diff
@@ -0,0 +1,118 @@
...
+  "/etc/nginx/conf.d/proxy.conf" :
+    mode: "000755"
+    owner: root
+    group: root
+    content: |
+      client_max_body_size  200M;
+      proxy_connect_timeout 1800;
+      proxy_send_timeout    1800;
+      proxy_read_timeout    1800;
+      send_timeout          1800;
```
mdibello: What's the difference between send_timeout and proxy_send_timeout? Does the former cover the latter or are they two distinct values?

.ebextensions/deploy.config
```diff
@@ -0,0 +1,118 @@
...
+  - option_name: BUNDLE_PATH
+    value: "vendor/bundle"
+
+packages:
+  yum:
+    git: []
+
+files:
+  "/etc/nginx/conf.d/proxy.conf" :
+    mode: "000755"
```
mdibello: For all of these files, you use permission mode 755 (-xwrx-rx-r). Is this the appropriate permission mode? Since the owning user and group are both root (and I'd assume the only user in group root is user root), root and group permissions should be the same, in theory. Also, who will be running these scripts? Write access only going to root makes sense, but should all users be able to execute these files?

.ebextensions/deploy.config
```diff
@@ -0,0 +1,118 @@
...
+      #!/usr/bin/env bash
+      if ! mount | grep /dev/xvdh > /dev/null; then
+        sudo mkfs -t ext3 /dev/sdh
+        sudo mkdir /media/volume1
+        sudo mount /dev/sdh /media/volume1
+      fi
+      sudo groupadd appdev
+      sudo usermod -aG appdev ec2-user
+      sudo usermod -aG appdev webapp
+      sudo chown -R webapp:appdev /media/volume1
```
mdibello: I think this line (48) is unnecessary because the same line appears below, and none of the lines between these two identical chown lines require the ownership change.

chikeabuah:
@mdibello i'm looking into everything you mentioned and making appropriate changes

Issue #2665, opened 2017-06-28
Warn users to check spam for our emails


doug-salvati:

doug-salvati:
PASSED testing on Chrome 58.0.3029.110 (linux).

Josh-Everett:
PASSED testing on Chrome 59.0.3071.115 (mac).

PR #2666, opened 2017-06-28
Add spam folder warnings when sending email

doug-salvati:
For #2665


doug-salvati:
Testers: these are just text changes, so no real need to test. If you want, create a user and make sure the flash message mentions spam

Issue #2686, opened 2017-08-09
Can't add tags directly on project page


doug-salvati:
Would be nice to be able to add tags with a "+" button right next to the tags list without having to go into Edit.
 
![image](https://user-images.githubusercontent.com/17438678/29136409-244c6e78-7d0b-11e7-9098-58594146415a.png)

PR #2687, opened 2017-08-10
Add/remove tags from projects page

doug-salvati:
For #2686


doug-salvati:
Meant to close this sooner

Issue #2748, opened 2017-11-08
Display bugs on contact page at certain widths


theyoyojo:
At width w on /contact:
(991 < w < 1200)

There are strange display issues with the grid. Example:

![isense-contact-ui-bug](https://user-images.githubusercontent.com/1909841/32562240-22dc8646-c47c-11e7-8376-828199cf92a5.png)

Josh-Everett:
PASSED testing on Safari 11.0.3 (mac).

Dayyvid:
FAILED testing on Chrome 64.0.3282.186 (windows).
On the dev server on http://dev.isenseproject.org/contact, Michael Kusmirek's info block is misaligned, as well as Farzeen's info block when screen is full size. On windowed mode, Michael's info block is still misaligned, but Farzeen's info block seems aligned.

Dayyvid:
![pt 1](https://user-images.githubusercontent.com/28739135/36865360-4fe7fd16-1d5d-11e8-9aac-3349a5283bf6.PNG)
![pt 2](https://user-images.githubusercontent.com/28739135/36865361-50c4c7f0-1d5d-11e8-8352-4903e5e71293.PNG)

doug-salvati:
Good catch. Thank you

PR #2777, opened 2017-12-06
[JS-2748] Fix display bug on contact page at certain widths

theyoyojo:
Widths: 991 < w < 1200


theyoyojo:
Resolves issue #2748

theyoyojo:
Because this is a minor fix at esoteric page widths that are not often used, I just fixed the grid layout. Some awkward display bugs remain, like the header of each box being inconsistently sized.

PR #2790, opened 2018-02-02
Inappropriate content reporting

theyoyojo:
DO NOT APPROVE THIS YET. The reporting emails are not correct.

Add report inappropriate content functionality.

Did not add dirty words to repo.

PR #2819, opened 2018-03-30
Missing end from merge conflict + adding test cases

doug-salvati:
When merging via github for #2790 , I took out an `end` keyword by mistake.


doug-salvati:
Unfortunately, now that the tests run all the way through we see that code coverage has dipped down.
@theyoyojo or myself must write some tests for the new code.

PR #2862, opened 2019-06-04
WIP: Dependency update

Codom:
Updating ruby dependencies and fixing all errors that come with these update.


Codom:
I have all tests passing, with no errors or warnings other than pending deprecation warnings from tests that still run. I want to confirm that this is the case on a couple of other machines before merging.

PR #2873, opened 2019-07-16
Hotfix/datetimepicker

Codom:
This may honestly not be merged in before (finally) releasing, as I realized that we should place higher priority on the color picker due to the current datetimepicker not having any issues. I'll keep this around to be functionally updated and merged later.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,7 +38,8 @@ group :development, :test do
 end
 
 # Jquery stuff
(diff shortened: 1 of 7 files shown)


Codom:
What doesn't work atm:
 * Timepicker
 * manual time specification (The date-picker will overwrite anything you type)
 * some tests need updating

